PR #7, opened 2020-05-27
Some suggestions

devmotion:
This PR contains a list of suggestions for `sinkhorn` and `sinkhorn2` (maybe one could use a bit more descriptive names? I guess they are motivated by the POT counterparts but it was completely unclear to me that one computes the map and the other one the cost without reading the documentation) that allow to use different array and number types such as CuArray and Float32. I also added some tests and some preliminary Travis integration (but I guess you have to activate it as well).

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -1,0 +1,22 @@
+# Documentation: http://docs.travis-ci.com/user/languages/julia/
+language: julia
+branches:
+  only:
+    - master
+os:
+  - linux
+  - osx
+julia:
+  - 1
+  - nightly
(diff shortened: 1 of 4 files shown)


zsteve:
Hi @devmotion, 

This is great! Thanks for the suggestions. Merging now.

> maybe one could use a bit more descriptive names? I guess they are motivated by the POT counterparts but it was completely unclear to me that one computes the map and the other one the cost without reading the documentation

Yes, that was the motivation for the naming. Although I agree with your comment, and could definitely change that.

PR #14, opened 2020-10-21
Proposing a Logo

davibarreira:
Created a logo for the package just for fun. Feel free to accept it in case you like it.


zsteve:
That is really nice! :) Do you think you would be able to set the logo text to "OptimalTransport.jl"? That way it refers to the package a little better. Otherwise I'm happy to merge

zsteve:
> That is really nice! :) Do you think you would be able to set the logo text to "OptimalTransport.jl"? That way it refers to the package a little better. Otherwise I'm happy to merge

Actually all good, just realised you had also the source .svg file there.

PR #15, opened 2020-11-23
Implementing Earth-Mover Distance in Julia

davibarreira:
I've implemented the `ead()` and `ead2()` functions in Julia, and changed the past functions to `pot_ead()` and `pot_ead2()`.
To implement these two functions, I've created auxiliary functions present in the "SimplexOT.jl" file.

These functions allow for 3 different methods (e.i. `ead(mu,nu,C, method)`) . The method "Native" correspond to my personal implementation of the Simplex method, using the North-West Corner described in the Computational Optimal Transport book.
This method works, but is not an efficient implementation of the Simplex algorithm, hence, it is there mostly for educational purposes.
The other two methods are "Simplex" and "IPM" (which stands for Interior Point Method). They require the packages JuMP.jl, Clp.jl and Tulip.jl, which were added to the dependencies.

The auxiliary function `ToSimplexFormat` adjustst the format of the OT problem to the standard Linear Programming format.

Also implemented tests for each of the functions.


devmotion:
It seems that would add quite many dependencies which is a bit unfortunate for me since I was hoping that it would be possible to reduce the number of big dependencies of OptimalTransport (https://github.com/zsteve/OptimalTransport.jl/issues/18). But it seems it is not necessary at all to depend on Jump, Tulip, or Clp. I would imagine one could just define the problem with MathOptInterface, and then users can just pick whatever solver they want to use.

I guess it could also be helpful to use sparse arrays (as I actually did when working with the northwest rule in https://github.com/devmotion/OptimalTransport.jl/blob/0e4fa870607ac974b8db57952b30bee1b8dd45be/src/earthmover.jl#L162-L199). I'm not sure how useful it is to include a naive implementation if it can't be used in practice - but maybe it could serve as the default fallback when going for a MathOptInterface formulation?

davibarreira:
I totally understand. I was reticent in adding JuMP and the others. My idea was to indeed use the Naive implementation, so everything would be native code. But the performance was way worst. I left it as perhaps someone (me included) could improve the implementation. Is "MathOptInterface" already a dependency? Perhaps using it would be indeed a better path. The North-West rule is just used for the Naive implementation. If it is discarded, the so can be the North-West Rule. Anyways, I'll check your implementation with sparse arrays.

davibarreira:
I just saw that you've implemented the Earth-Movers distance. Any reason why this isn't added to OptimalTransport.jl?

devmotion:
There was some problem with it (Stephen opened an issue but I was busy and couldn't look into it) and it's also not the most performant implementation (it is just a simple implementation of the explanations in the Computational Optimal Transport book with the help of LightGraphs).

devmotion:
And it existed before this package was created, I guess that's why it was not included here in the first place.

davibarreira:
I see. I'll take a look at how to actually perform the Simplex with MathOptInterface instead of JuMP and leave the choice of Optimizer for the user. But as you've pointed out, it would be good to have a native implementation  to fall back on. Yet, mine might be too rough on the edges, but perhaps after some polishing it might become useful.

devmotion:
>  I'm supposing they are a more efficient way to store sparse matrices. Is that correct?

Yes, exactly. They can be more efficient if you many entries are zero.

devmotion:
But actually I think it's probably not a good idea. I don't think one can generally assume that the transport plan are so sparse (< 0.01% or so) that SparseArrays are more efficient.

davibarreira:
Thanks for the heads up. I’ll then focus on implementing the optimization in MathOptInterface. Then, I’ll see how I can improve on my Simplex implementation. I might end up ditching it, and leaving only the optimized version.

davibarreira:
So, I'm reading a bit about MathOptInterface, and I saw some people advising to use JuMP, saying that it actually wasn't much heavier. What do you think, @devmotion? Should I keep using JuMP and just remove the dependencies on  the optimizers? @zsteve what do you think? Also, I'll just getting rid of my Simplex implementation and the North-West method.

devmotion:
I'll explain more clearly what I had in mind: For both solvers, you just rewrite the optimal transport problem as a LP problem in exactly the same way. The main point is the formulation of the problem and not which solver is used to solve it in the end, Clp or Tulip are just two possible solvers. So I don't think it is useful to have an implementation for two specific solvers and to depend on them. Instead in my opinion it would be more useful to allow users to specify a solver as one of the arguments of `emd` and `emd2`, and at most to depend on one default solver (however, I'm not in favour of this, I think users should just load Clp or Tulip or whatever solver they want to use). Now the question is, how can the LP problem be formulated in a generic way that all the solvers understand? As far as I understand, that's exactly the purpose of MathOptInterface which is supported by, e.g., Clp and Tulip: it provides a general interface for defining these problems and a standard API for solving them that is implemented by the different solvers. I don't think there would be an advantage in using the DSL of Jump instead of MathOptInterface. Jump has additional dependencies such as ForwardDiff, DataStructures, and Calculus.

davibarreira:
I totally agree with you about the Optimizers. It doesn't make sense to have a function for each, you are correct. I haven't actually worked with MathOptInterface, only with JuMP. But, the reason I was favoring using JuMP was because of things I read like

"Due to some restrictions in Clp's C api, the Clp's MathOptInterface wrapper does not support directly modifying a problem after it has been created, e.g., changing variable bounds or modifying constraints coefficients.

Therefore, we highly recommend that you use the Clp.jl package with higher-level package such as JuMP.jl."

But perhaps you are right in worrying about the extra dependencies. I'll try implement the problem directly with MathOptInterface.

davibarreira:
Done. I've just submitted a new pull request with ead implemented on MathOptInterface, leaving up  to the user to choose the optimizer.

zsteve:
> But actually I think it's probably not a good idea. I don't think one can generally assume that the transport plan are so sparse (< 0.01% or so) that SparseArrays are more efficient.

I think for the Earth-Mover distance typically there are O(n) nonzero entries, so for sufficiently large problems the transport plan would be sparse. Still, I'm not sure if using sparse arrays by default is the best way to go. Well, I guess I"ll take a look at the most recent pull request to see what you've decided to go with ;)

Regarding optimizer dependencies, I agree that the solver should be generic. MathOptInterface sounds like it should do the job! 

About a truly native implementation of the EMD algorithm in Julia, I looked into this a while ago as David noted and it seems quite an involved task to come up with a performant implementation -- the Python-OT package uses a custom piece of C++ code that seems to have been handed down from researcher to researcher...
So I think using LP solvers should do the job for now. Btw, is there any comparison between the LP solver-based implementation vs. the Python OT (C++) code?

davibarreira:
I've used the Python implementation to test the new functions. But I haven't compared them in terms of performance, only checked if the output was the same. I ended up not implementing with sparse arrays, and in the end I removed the native simplex solver, sincr it was much slower.

PR #20, opened 2020-11-30
Update CI

devmotion:
This PR tries to address https://github.com/zsteve/OptimalTransport.jl/issues/18#issuecomment-735473875.

It:
- sets up CI with Github Actions for Linux, MacOS, and Windows (it worked for my package and recently Travis introduced quite strict limits even for public projects)
- adds TagBot
- updates CompatHelper
- builds documentation

Note that you have to add a deploy key to the repo such that CompatHelper is allowed to run tests and TagBot is allowed to rebuild the documentation on a new release: https://juliadocs.github.io/Documenter.jl/stable/man/hosting/#Authentication:-SSH-Deploy-Keys


devmotion:
It is unfortunate that the package states that it is compatible with Julia 1.0 but it can't be tested on Julia 1.0 since CUDA is not available for Julia 1.0. Maybe it would be good to separate the CUDA tests and only run them on more recent Julia versions or just drop support for Julia < 1.3. For now I only removed tests with Julia 1.0.

PR #30, opened 2020-12-12
Fix LP formulation and solver

devmotion:
Unfortunately, there are more problems with the merged LP formulation and solver than just the unnecessary dependencies.

Apart from the incorrect formatting (docstrings, mix of tabs and whitespace, and code) the implementation was unnecessarily complex and partly incorrect (the dimensions of the optimal transport map and the LP formulation were not consistent with the docstring). It is not necessary to add zero padding in the LP formulation, and in fact it is not even needed to allocate any new vectors or matrices. Also POT does not require histograms of the same size.

This PR fixes all of these issues.


zsteve:
Hi @devmotion,
Sorry this slipped past me, and thanks for picking up on it.
I ran a couple of my own tests before merging the LP code to check a few cases but didn't look closely at the actual code. 
Stephen

PR #31, opened 2020-12-13
Fix compat bounds

devmotion:
Compat bounds have to be upper bounded otherwise a new version won't be made available in the Julia package registry. One has to list the supported versions explicitly if they are breaking according to the Julia semver system (BTW is there a reason to support LazyArrays 0.18 as well?).


zsteve:
Sorry about that, was trying to pull the updated package to my local machine after the series of merges and ran into a compat error, and wondered if that could've been the issue. Turns out it's not, so all good.

Thanks for the tip though, will keep in mind.

PR #45, opened 2021-05-14
1d Optimal Transport

davibarreira:
Implementation of the **1D version of Optimal Transport for cost functions of the form `c(x,y) = h(|x-y|)` where `h`is convex**.
# Files Modified
1) `ot1d.jl` - has the 1D implementations;
2) `OptimalTransport.jl` - imported ot1d.jl and the packages `Distributions.jl` and `QuadGK.jl` which are the new dependencies added (actually, Distributions.jl already requires QuadGK.jl);
3) `runtests.jl`- added tests for the 1D implementations;
4) `1D - Optimal Transport.ipynb` - notebook showcasing the use of the new functions.

# Functions implemented
There were 3 main functions implemented, the `otCost1d`, `otPlan1d` and `otCostPlan1d`, where the first returns the cost,
the second returns the plan, and the third returns both. Each of these functions has 3 multiple dispatch implementations.
First:
* function otCost1d(c,μ::Distributions.UnivariateDistribution,ν::Distributions.UnivariateDistribution)
* function otPlan1d(c,μ::Distributions.UnivariateDistribution,ν::Distributions.UnivariateDistribution)
* function otCostPlan1d(c,μ::Distributions.UnivariateDistribution,ν::Distributions.UnivariateDistribution)
These functions allows the user to calculate the distance between continuous distributions from the Distributions.jl.

Second:
* function otPlan1d(c,u::Vector,u_weights::Vector,v::Vector,v_weights::Vector)
* function otCost1d(c,u::Vector,u_weights::Vector,v::Vector,v_weights::Vector)
* function otCostPlan1d(c,u::Vector,u_weights::Vector,v::Vector,v_weights::Vector)
These implement the discrete algorithm when the user provides the vectors containing the position of the Dirac masses and
the vector of weights.

Third:
* function otCost1d(c,μ::Distributions.DiscreteNonParametric,ν::Distributions.DiscreteNonParametric)
* function otPlan1d(c,μ::Distributions.DiscreteNonParametric,ν::Distributions.DiscreteNonParametric)
* function otCostPlan1d(c,μ::Distributions.DiscreteNonParametric,ν::Distributions.DiscreteNonParametric)
These implement the discrete algorithm when the user provides a DiscreteNonParametric distribution, which is the way
`Distributions.jl` allows the user to define arbitrary finite discrete distributions. Just note that when one uses this, the distribution is automatically sorted, so the transport plan will correspond to the sorted distribution.


src/ot1d.jl
```diff
@@ -0,0 +1,283 @@
+"""
+otCost1d(c,μ::Distributions.UnivariateDistribution,
+           ν::Distributions.UnivariateDistribution)
```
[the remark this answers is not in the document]
devmotion: Maybe one could also rename `c` to `cost` (at least in the docstring) to make it even clearer.

src/ot1d.jl
```diff
@@ -0,0 +1,283 @@
...
+Parameters:\n
+c: The cost function, which should be of the form ``c(x,y) = h(abs(x-y))``
+where `h` is a convex function.
+μ: 1-D distribution (e.g. `μ = Distributions.Normal(0,1)`)\n
+ν: 1-D distribution (e.g. `μ = Distributions.Normal(0,1)`)\n
+"""
+function otCost1d(c,μ::Distributions.UnivariateDistribution,ν::Distributions.UnivariateDistribution)
+    g(μ,ν,x) = c(quantile(μ,x),quantile(ν,x))
+    f(x) = g(μ,ν,x)
+    return quadgk(f,0,1)[1]
```
devmotion, line 19: It is not necessary to create all these functions:
```suggestion
    integral, _ = quadgk(0, 1) do x
        return c(quantile(μ, x), quantile(ν, x))
    end
    return integral
```

src/ot1d.jl
```diff
@@ -0,0 +1,283 @@
...
+end
+
+"""
+otPlan1d(c,μ::Distributions.UnivariateDistribution,
+            ν::Distributions.UnivariateDistribution)
```
devmotion, line 24: Again style:
```suggestion
    otPlan1d(c, μ::UnivariateDistribution, ν::UnivariateDistribution)
```

src/ot1d.jl
```diff
@@ -0,0 +1,283 @@
...
+
+Calculates the Optimal Transport Plan between μ to ν, where
+they are 1-Dimensional distributions and the cost
+function is of the form ``c(x,y) = h(|x-y|)`` such that
+``h`` is a convex function.
```
devmotion, line 29: Style:
```suggestion
Calculate the optimal transport map of the Monge-Kantorovich problem with univariate distributions `μ` and `ν` as source and target marginals and cost function `c`.
Cost function `c` has to be of the form ``c(x, y) = h(|x - y|)`` where ``h`` is a
convex function.
```

src/ot1d.jl
```diff
@@ -0,0 +1,283 @@
...
+
+Parameters:\n
+c: The cost function, which should be of the form ``c(x,y) = h(|x-y|)``
+where ``h`` is a convex function.
+μ: 1-D distribution (e.g. `μ = Distributions.Normal(0,1)`)\n
+ν: 1-D distribution (e.g. `μ = Distributions.Normal(0,1)`)\n
```
devmotion, line 35: Again this should be omitted:
```suggestion
```

src/ot1d.jl
```diff
@@ -0,0 +1,283 @@
...
+
+Returns the Optimal Transport Plan as a function
+  ```math
+  T(x)=F_\\nu^{-1}(F_\\mu(x)).
+  ```
```
devmotion, line 40: Style:
````suggestion
The optimal transport map ``T`` is returned as a function. It is defined as
```math
T(x) = F_{\\nu}^{-1}(F_{\\mu}(x)),
```
where ``F_{\\mu}`` and ``F_{\\nu}`` are the cumulative distribution functions of
distributions ``\\mu`` and ``\\nu``, respectively.
````
Maybe one should also rename `T` here to `\gamma` to make the notion consistent throughout the package:

````suggestion
The optimal transport map ``\\gamma`` is returned as a function. It is defined as
```math
\\gamma(x) = F_{\\nu}^{-1}(F_{\\mu}(x)),
```
where ``F_{\\mu}`` and ``F_{\\nu}`` are the cumulative distribution functions of
distributions ``\\mu`` and ``\\nu``, respectively.
````

File: src/ot1d.jl [hunk 7]
```diff
@@ -0,0 +1,283 @@
...
+"""
+function otPlan1d(c,μ::Distributions.UnivariateDistribution,ν::Distributions.UnivariateDistribution)
```
devmotion, line 42:
```suggestion
function otPlan1d(c, μ::UnivariateDistribution, ν::UnivariateDistribution)
```

src/ot1d.jl
```diff
@@ -0,0 +1,283 @@
...
+  T(x) = Distributions.quantile(ν,Distributions.cdf(μ,x))
+  return T
```
devmotion, line 44:
```suggestion
  T(x) = quantile(ν, cdf(μ, x))
  return T
```

src/ot1d.jl
```diff
@@ -0,0 +1,283 @@
...
+  ```math
+  T(x)=F_\\nu^{-1}(F_\\mu(x)).
+  ```
+"""
+function otCostPlan1d(c,μ::Distributions.UnivariateDistribution,ν::Distributions.UnivariateDistribution)
+    g(μ,ν,x) = c(quantile(μ,x),quantile(ν,x))
+    f(x) = g(μ,ν,x)
+    T(x) = Distributions.quantile(ν,Distributions.cdf(μ,x))
+    return quadgk(f,0,1)[1], T
+end
```
devmotion, line 73: I don't think this function should be added since it is not more efficient than just computing the cost and map separately.
```suggestion
```
In any case, all suggestions from above apply here as well.

src/ot1d.jl
```diff
@@ -0,0 +1,283 @@
...
+v        : Vector of point mass location    (e.g. `v = collect(10:20)`)\n
+v_weights: Vector of weights for each point (e.g. `v_weights = ones(10)*0.1`)\n
+
+
+For the second case we have parameters:\n
+c: The cost function, which should be of the form ``c(x,y) = h(abs(x-y))``
+where ``h`` is a convex function. \n
+μ: Finite Discrete Distribution (e.g. `μ = DiscreteNonParametric(u, u_n)`)\n
+ν: Finite Discrete Distribution (e.g. `ν = DiscreteNonParametric(v, v_n)`)\n
+"""
```
devmotion, line 98: Same style issues as above. Additionally, the docstring for `DiscreteNonParametric` should be moved to the corresponding function.

src/ot1d.jl
```diff
@@ -0,0 +1,283 @@
...
+function otCost1d(c,u::Vector,u_weights::Vector,v::Vector,v_weights::Vector)
```
devmotion, line 99: Maybe one should only accept `DiscreteNonParametric` (and no vectors) and require users to construct it from the support and probabilities.

This would be more consistent with the other methods that only accept cost and marginals, and the support would already be sorted (`DiscreteNonParametric` does that automatically).

src/ot1d.jl
```diff
@@ -0,0 +1,283 @@
...
+
+  @assert length(u) == length(u_weights)
+  @assert length(v) == length(v_weights)
+  @assert sum(u_weights) ≈ 1.0  atol=1e-7
+  @assert sum(v_weights) ≈ 1.0  atol=1e-7
```
devmotion, line 104: `@assert` should not be used for checking inputs.

src/ot1d.jl
```diff
@@ -0,0 +1,283 @@
...
+
+  cost = 0
```
devmotion, line 106: This creates a type instability.

src/ot1d.jl
```diff
@@ -0,0 +1,283 @@
...
+  perm_u = sortperm(u)
+  perm_v = sortperm(v)
+  u_sorted = u[perm_u]
+  v_sorted = v[perm_v]
+  u_w_sorted = u_weights[perm_u]
+  v_w_sorted = v_weights[perm_v]
```
devmotion, line 112: This can be removed if only `DiscreteNonParametric` is accepted:

```suggestion
  perm_u = sortperm(u)
  perm_v = sortperm(v)
  u_sorted = u[perm_u]
  v_sorted = v[perm_v]
  u_w_sorted = u_weights[perm_u]
  v_w_sorted = v_weights[perm_v]
```

src/ot1d.jl
```diff
@@ -0,0 +1,283 @@
...
+          if j == length(v_sorted)+1
+              break
+          end
+          wi -= wj
+          wj  = v_w_sorted[j]
+      end
+  end
+  
+  return γ
+end
```
devmotion, line 200: Same comments as above. Ideally, one would also not copy all the code from above.

src/ot1d.jl
```diff
@@ -0,0 +1,283 @@
...
+          if j == length(v_sorted)+1
+              break
+          end
+          wi -= wj
+          wj  = v_w_sorted[j]
+      end
+  end
+  
+  return cost, γ
+end
```
devmotion, line 268: Same comments, and ideally the code should be refactored such that it is not duplicated.

devmotion: More generally, again I don't see the need for a method that computes both things at the same time. Here one can just compute the map and then compute the cost afterwards if desired, similar to `emd2`. This function is not more efficient than the "naive" approach.

src/ot1d.jl
```diff
@@ -0,0 +1,283 @@
...
+
+function otCost1d(c,μ::Distributions.DiscreteNonParametric,
+                    ν::Distributions.DiscreteNonParametric)
```
devmotion, line 271: As mentioned, I think one should *only* accept DiscreteNonParametric, no vectors.

test/runtests.jl
```diff
@@ -43,6 +44,57 @@ Random.seed!(100)
...
+
+    μ = rand(n)
+    ν = rand(m) .+ 0.5;
+    μ_n = rand(n)
+    ν_m = rand(m)
+    μ_n = μ_n/sum(μ_n)
+    ν_m = ν_m/sum(ν_m);
+
+    c(x,y) = (x-y)^2
+    C = Distances.pairwise(Distances.SqEuclidean(), μ', ν');
```
devmotion, line 66:
```suggestion
    C = pairwise(SqEuclidean(), μ', ν');
```

devmotion (changes requested):
I added some comments. Can you also sync the PR with the master branch? It seems you worked with an outdated version of OptimalTransport.

It would also be interesting to see some benchmarks with the Tulip solver. The exact algorithm should be more efficient in the new version but probably there is still quite some overhead.

docs/src/index.md
```diff
@@ -7,6 +7,28 @@ emd
 emd2
 ```
 
+### Exact optimal transport for 1D.
+
+```@docs
+otCost1D
+otPlan1D
+otCostPlan1D
```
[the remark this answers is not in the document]
davibarreira: Thanks for all the comments. I'm a newbie in package dev, so it's great to get direct feedback. I should've read the style guide.

When you say "Can you also sync the PR with the master branch?". Once the pull request is merged, wouldn't it merge with the current master branch? If not, should I merge the PR in my machine and then resubmit it?

I think we should use `otplan` and `otcost`. It's more explicit than `emd` and `emd2` in my opinion, while not as verbose as `optimal_transport_plan` and `optimal_transport_cost`.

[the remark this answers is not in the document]
davibarreira: About the existence of an specific function for 1D, I'm shook about it.
POT.py has the 1D version explicitly. I'd guess the reason is due to SlicedWasserstein, which uses the 1D version, so making the function explicit would guarantee that you are working in the correct dimension. 
I guess, due to the fact that 1D Optimal Transport is in general so distinctive to the larger dimensions, that perhaps it should indeed have an specific function just for it. But I don't really have a strong opinion about it.

docs/src/index.md
```diff
@@ -7,6 +7,28 @@ emd
...
+```
+
+For a cost function ``c(x,y) = h(|x-y|)`` where
+The Optimal Transport Cost for 1D distributions can be obtained by:
+```math
+\\int_0^1 h(F_\\nu^{-1}(y) - F_\\mu^{-1}(y)) dy
+```
+Note that for the p-Wasserstein, we obtain:
```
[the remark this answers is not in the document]
davibarreira: Yeah, I think we can just put on the docstring. By the way, this was a typo. I forgot to delete this last line.

src/ot1d.jl
davibarreira: We could follow POT.py, and actually implement one function that calculates the plan and the cost behind the scenes, while
the `otcost1d` returns the first argument and `otplan1d` the second.

src/ot1d.jl
davibarreira: I thought about, and the reason I wrote this was the following. First, as a user, I'd like a function that returned both, freeing me of the need to get the plan and calculate the cost myself. Secondly, depending on your objective, you might be interested in both the plan and the cost. Now, if the input is the 1D, without a function that returns both, the user would either need to run both `otcost1d` and `otplan1d`, which would be inefficient. The other option would be to run `otplan1d`, and to calculate the cost explicitly. But then, the user would need to write a function to calculate such cost, which is not as direct as `dot(C,P)`, since the plan and the cost might be a functions, and not matrices.

src/ot1d.jl
davibarreira: I don't have objections to this. If you think it's better, than we can go with that.

src/ot1d.jl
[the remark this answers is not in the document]
davibarreira: I avoided doing this since I usually call `cost` a value, leaving `c` to be a function.

examples/1D - Optimal Transport.ipynb
```diff
@@ -0,0 +1,5883 @@
+{
```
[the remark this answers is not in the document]
davibarreira: I like using notebooks, because they are quick to make and easy to follow, specially if you are trying quickly understand how to use a package (it really helps me when repos have notebooks which I can just go to and follow along). I do agree that they are not well integrated with the documentation, but my suggestion would be to create examples in the documentation as you pointed out, AND leave a set o notebooks in the repo, so people can just go through them.
If you thinks this could get too messy, than I can just create another repository with Optimal Transport usage examples, and we can link to it in the README. I'm also gonna start a repo for generating OT plots.

src/ot1d.jl
davibarreira: Sure, I agree that this would add consistency. But do you think we should also require this in the other functions such as `emd` and `emd2`?

examples/1D - Optimal Transport.ipynb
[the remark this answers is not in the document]
devmotion: I would very much prefer a Literate example. They are Julia scripts with comments that can be exported to notebooks automatically (which can be included in the gh-pages branch) and included in the documentation, as the examples above demonstrate. And the additional advantage is that they are just regular Julia scripts so one can also just run them in the REPL without setting up a Jupyter environment first.

Project.toml
```diff
@@ -5,9 +5,12 @@ version = "0.2.1"
 
 [deps]
 Distances = "b4f34e82-e78d-54a5-968a-f98e89d6e8f7"
+Distributions = "31c24e10-a181-5473-b8eb-7969acd0382f"
+FillArrays = "1a297f60-69ca-5386-bcde-b61e274b549b"
```
devmotion, line 9: This seems wrong and probably should not be reintroduced:
```suggestion
```

devmotion:
My high-level suggestion here similar to the other PR (I'll add a more detailed review later):
- rename and deprecate `emd` to `optimal_transport_plan`, and change the order of the arguments to `C, mu, nu` for consistency
- rename and deprecate `emd2` to `optimal_transport_cost`, and change the order of the arguments to `C, mu, nu` for consistency
- add a keyword argument `plan` to `optimal_transport_cost` that allows to specify the optimal transport plan if already computed by the user (similar to the keyword argument `mean` in `std` and `var`)
- rename the functions in this PR accordingly and remove the functions that compute both the plan and the cost
- remove the functions that take both the support and probabilities as inputs and only keep the versions with DiscreteNonParametric

davibarreira:
For me this all sounds good. But I have one question about the computation of plan and cost in the same function. If we don't have a function that returns both at once, how do we make sure we efficiently calculate the plan and the cost without repeating some of the calculations? For example, the function for the Normals returns a map `T`, instead of a matrix. How would this be used to avoid repeating the computations when calculating the cost?

devmotion:
> For example, the function for the Normals returns a map T, instead of a matrix. How would this be used to avoid repeating the computations when calculating the cost?

It depends on the specific use case. If it is more efficient to compute the cost independently, the keyword argument (probably a default of `nothing` would be good) would not be supported (probably preferable) or just not be used (this should be mentioned in the docstring). In general, it does not matter if the plan is a function or a matrix for the keyword argument functionality.

E.g., in the case of continuous univariate distributions one could use `T(x) = quantile(nu, cdf(mu, x))`, if it is provided, to compute the cost according to
```julia
quadgk(0, 1) do q
    x = quantile(mu, q)
    y = T(x)
    return c(x, T(x))
end
```
This will lead the same result as computing the cost from scratch according to
```julia
quadgk(0, 1) do q
    return c(quantile(mu, q), quantile(nu, q))
end
```
which would be preferable if `T` is not provided.

src/ot1d.jl
```diff
@@ -0,0 +1,245 @@
+"""
+    otCost1d(c, μ::UnivariateDistribution, ν::UnivariateDistribution)
+
+Calculate the optimal transport cost of the Monge-Kantorovich problem with univariate distributions `μ` and `ν` as source and target marginals and cost function `c`.
+
+Cost function `c` has to be of the form ``c(x, y) = h(|x - y|)`` where ``h`` is a
+convex function.
+"""
+function otCost1d(c, μ::UnivariateDistribution, ν::UnivariateDistribution)
```
devmotion, line 9: BTW this should be restricted to
```suggestion
function otCost1d(
    c, μ::ContinuousUnivariateDistribution, ν::ContinuousUnivariateDistribution
)
```

davibarreira: Why restrict to Continuous? This still works for cases such as a Continuous Distributions and a Discrete Distribution.

davibarreira: Also, if the distributions is infinitely discrete, the other code would not work, but this one would. So I think we should just leave as it was.

devmotion:
No, this would not be type stable. I suggested the same design as `std` and `var` in the Statistics standard library: it has a keyword argument `mean` that can be used to specify a precomputed `mean` that increases the efficiency of the computation of `std` and `var` (since it does not have to be recomputed). In our case this would mean we have
```julia
function optimal_transport_map(c, mu, nu)
    ...
end
```
and
```julia
function optimal_transport_cost(c, mu, nu; map=nothing)
    ...
end
```
If users have already computed the optimal transport map, they can provide it as a keyword argument when computing the cost which possibly speeds up the computation (since the transport map does not have to be recomputed).

examples/1D - Optimal Transport.ipynb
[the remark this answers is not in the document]
devmotion: I updated the outdated example in https://github.com/zsteve/OptimalTransport.jl/pull/47 and moved it to the suggested setup. The example can be viewed here: https://zsteve.github.io/OptimalTransport.jl/previews/PR47/examples/basic/ The Jupyter notebook can be viewed here: https://nbviewer.jupyter.org/github/zsteve/OptimalTransport.jl/blob/gh-pages/previews/PR47/examples/basic.ipynb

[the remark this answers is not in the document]
devmotion: This shows once more that there is really no need for adding notebooks to the repo even if you want to provide Jupyter notebooks for users.

Regarding plots, I think it might be best to just provide simple recipes with RecipesBase in OptimalTransport.jl.

src/ot1d.jl
devmotion: In general, `otPlan1D(c, ::UnivariateDistribution, ::UnivariateDistribution)` is incorrect for finite discrete distributions and also semi-discrete optimal transport since the optimal transport map is not deterministic anymore and masses have to be split. So it seems wrong to define it for distributions that are not continuous, doesn't it?

davibarreira: Actually, I think it should be `otplan1d(c, ::ContinuousDistribution, ::UniveriateDistribution)`. The only condition for the the map to exists it that `u << Lebesgue` and `c(x,y) = h(|x-y|)` for `h` convex.

davibarreira:
What is missing for this PR to be accepted? Any other changes I should make?

devmotion:
Could you address the comments in the first review or add a comment to the suggestions? Then I'll do a second round of review.

davibarreira:
> No, this would not be type stable. I suggested the same design as `std` and `var` in the Statistics standard library: it has a keyword argument `mean` that can be used to specify a precomputed `mean` that increases the efficiency of the computation of `std` and `var` (since it does not have to be recomputed). In our case this would mean we have
> 
> ```julia
> function optimal_transport_map(c, mu, nu)
>     ...
> end
> ```
> 
> and
> 
> ```julia
> function optimal_transport_cost(c, mu, nu; map=nothing)
>     ...
> end
> ```
> 
> If users have already computed the optimal transport map, they can provide it as a keyword argument when computing the cost which possibly speeds up the computation (since the transport map does not have to be recomputed).


I'm fixing the PR and I was thinking about this suggestion. Why not instead use multiple dispatch? e.g.
```
optimal_transport_cost(C,mu,nu,plan)
optimal_transport_cost(C,mu,nu)
```

devmotion:
> Why not instead use multiple dispatch?

There are at least two reasons, I think: it is more convenient and clearer with keyword arguments (there's only one function syntax that users have to know and it is clear what the name of the keyword argument clearly states what it is supposed to be which is not clear if it is just the fourth positional argument) and in general we do not want to dispatch based on the type of the plan.

Internally, though it can be convenient sometimes (not always) to forward some keyword arguments to an internal function that allows to dispatch on the type (that's common practice also in base Julia or Statistics). I.e., one would define `f(one_arg, two_arg; name=some_keyword) = _f(one_arg, two_arg, some_keyword)` and then define implementations for `_f`. So the keyword argument approach is not limiting and one can still dispatch at some stage if desired.

davibarreira:
> > Why not instead use multiple dispatch?
> 
> There are at least two reasons, I think: it is more convenient and clearer with keyword arguments (there's only one function syntax that users have to know and it is clear what the name of the keyword argument clearly states what it is supposed to be which is not clear if it is just the fourth positional argument) and in general we do not want to dispatch based on the type of the plan.
> 
> Internally, though it can be convenient sometimes (not always) to forward some keyword arguments to an internal function that allows to dispatch on the type (that's common practice also in base Julia or Statistics). I.e., one would define `f(one_arg, two_arg; name=some_keyword) = _f(one_arg, two_arg, some_keyword)` and then define implementations for `_f`. So the keyword argument approach is not limiting and one can still dispatch at some stage if desired.

If instead one used `optimal_transport_cost(C,mu,nu;plan)`, wouldn't it be similar to `optimal_transport_cost(C,mu,nu;plan=nothing)`?

devmotion:
> If instead one used multiple dispatch `optimal_transport_cost(C,mu,nu;plan)`, wouldn't it be similar to `optimal_transport_cost(C,mu,nu;plan=nothing)`?

Hmm it's not clear to me what exactly you are asking here. If you just define `f(arg; kw)` then it will error if users don't provide a value for `kw` whereas if you define `f(arg; kw=nothing)` then `f(arg)` will call `f(arg; kw=nothing)`.

davibarreira:
> > If instead one used multiple dispatch `optimal_transport_cost(C,mu,nu;plan)`, wouldn't it be similar to `optimal_transport_cost(C,mu,nu;plan=nothing)`?
> 
> Hmm it's not clear to me what exactly you are asking here. If you just define `f(arg; kw)` then it will error if users don't provide a value for `kw` whereas if you define `f(arg; kw=nothing)` then `f(arg)` will call `f(arg; kw=nothing)`.

I see. I thought multiple dispatch worked with key-word arguments.

src/ot1d.jl
davibarreira: What do you recommend?

davibarreira:
I've done all the modifications suggested (I think). I've removed the jupyter notebook. I'll later learn how to properly use Literate.jl and write an example of use in the documentation.

src/OptimalTransport.jl
```diff
@@ -672,4 +675,162 @@ function quadreg(mu, nu, C, ϵ; θ=0.1, tol=1e-5, maxiter=50, κ=0.5, δ=1e-5)
...
+        g(μ, ν, x) = c(quantile(μ, x), quantile(ν, x))
+        f(x) = g(μ, ν, x)
+        return quadgk(f, 0, 1)[1]
+    else
+        cost = quadgk(0, 1) do q
+            x = quantile(μ, q)
+            return c(x, plan(x))
+        end
+        return cost[1]
+    end
```
devmotion, line 709: There is no need to define `g` and in particular `f` explicitly here:
```suggestion
    cost, _ = if plan === nothing
        quadgk(0, 1) do q
            return c(quantile(μ, q), quantile(ν, q))
        end
    else
        quadgk(0, 1) do q
            x = quantile(μ, q)
            return c(x, plan(x))
        end
    end
    return cost
```

src/OptimalTransport.jl
```diff
@@ -672,4 +675,162 @@ function quadreg(mu, nu, C, ϵ; θ=0.1, tol=1e-5, maxiter=50, κ=0.5, δ=1e-5)
...
+Parameters:\n
+c: The cost function, which should be of the form ``c(x,y) = h(|x-y|)``
+where ``h`` is a convex function.
+μ: 1-D distribution (e.g. `μ = Distributions.Normal(0,1)`)\n
+ν: 1-D distribution (e.g. `μ = Distributions.Normal(0,1)`)\n
+
+Returns the optimal transport plan as a function
+  ```math
+  T(x)=F_\\nu^{-1}(F_\\mu(x)).
+  ```
```
devmotion, line 729: Same here, the docstring should be improved.

src/OptimalTransport.jl
```diff
@@ -672,4 +675,162 @@ function quadreg(mu, nu, C, ϵ; θ=0.1, tol=1e-5, maxiter=50, κ=0.5, δ=1e-5)
...
+"""
+function optimal_transport_plan(
+    c, μ::ContinuousUnivariateDistribution, ν::UnivariateDistribution
+)
+    # Use T instead of γ to indicate that this is a Monge map.
+    T(x) = Distributions.quantile(ν, Distributions.cdf(μ, x))
```
devmotion, line 735: No need to qualify the functions:
```suggestion
    T(x) = quantile(ν, cdf(μ, x))
```

src/OptimalTransport.jl
```diff
@@ -672,4 +675,162 @@ function quadreg(mu, nu, C, ϵ; θ=0.1, tol=1e-5, maxiter=50, κ=0.5, δ=1e-5)
...
+``h`` is a convex function.
+
+Parameters:\n
+c: The cost function, which should be of the form ``c(x,y) = h(abs(x-y))``
+where ``h`` is a convex function. \n
+μ: Finite Discrete Distribution (e.g. `μ = DiscreteNonParametric(u, u_n)`)\n
+ν: Finite Discrete Distribution (e.g. `ν = DiscreteNonParametric(v, v_n)`)\n
+
+Returns the cost.
+"""
```
devmotion, line 754: Again the docstring should be fixed. Seems it could even be removed, since there is any additional information here?
```suggestion
```

src/OptimalTransport.jl
```diff
@@ -672,4 +675,162 @@ function quadreg(mu, nu, C, ϵ; θ=0.1, tol=1e-5, maxiter=50, κ=0.5, δ=1e-5)
...
+function optimal_transport_cost(
+    c, μ::DiscreteNonParametric, ν::DiscreteNonParametric; plan=nothing
+)
+    if plan === nothing
+        return _optimal_transport_cost_plan(c, μ, ν)[1]
+    else
+        return dot(StatsBase.pairwise(c, μ.support, ν.support), plan)
+    end
+
+    return cost
```
devmotion, line 764: There are some possible improvements here:
- it is not necessary to compute the plan in the first branch
- it is not necessary to compute the matrix of pairwise distances explicitly in the second case
- `return cost` is never reached

```suggestion
    cost = if plan === nothing
        _optimal_transport_cost_plan(c, μ, ν)[1]
    else
        dot(StatsBase.pairwise(c, μ.support, ν.support), plan)
    end
    return cost
```

src/OptimalTransport.jl
```diff
@@ -672,4 +675,162 @@ function quadreg(mu, nu, C, ϵ; θ=0.1, tol=1e-5, maxiter=50, κ=0.5, δ=1e-5)
...
+``h`` is a convex function.
+
+Parameters:\n
+c: The cost function, which should be of the form ``c(x,y) = h(abs(x-y))``
+where ``h`` is a convex function. \n
+μ: Finite Discrete Distribution (e.g. `μ = DiscreteNonParametric(u, u_n)`)\n
+ν: Finite Discrete Distribution (e.g. `ν = DiscreteNonParametric(v, v_n)`)\n
+
+Returns the optimal transport plan γ as a matrix.
+"""
```
devmotion, line 782: Again it seems the docstring should be removed?
```suggestion
```

src/OptimalTransport.jl
```diff
@@ -672,4 +675,162 @@ function quadreg(mu, nu, C, ϵ; θ=0.1, tol=1e-5, maxiter=50, κ=0.5, δ=1e-5)
...
+function optimal_transport_plan(
+    c, μ::DiscreteNonParametric, ν::DiscreteNonParametric; plan=nothing
+)
+    return _optimal_transport_cost_plan(c, μ, ν)[2]
```
devmotion, line 786: It is a bit suboptimal to compute the cost as well here.

src/OptimalTransport.jl
```diff
@@ -672,4 +675,162 @@ function quadreg(mu, nu, C, ϵ; θ=0.1, tol=1e-5, maxiter=50, κ=0.5, δ=1e-5)
...
+)
+
+Auxiliary function that calculates the cost and plan for
+μ an ν where both  they are 1-Dimensional discrete distributions with finite support
+and the cost function is of the form ``c(x,y) = h(|x-y|)`` such that
+``h`` is a convex function.
+
+Returns cost and γ, where cost represtents the optimal transport cost and
+γ is the optimal transport plan given as a matrix.
+"""
```
devmotion, line 801: It's not necessary to add docstrings for internal methods but it's great if there is some documentation :+1: However, as above the format should be fixed.

src/OptimalTransport.jl
```diff
@@ -672,4 +675,162 @@ function quadreg(mu, nu, C, ϵ; θ=0.1, tol=1e-5, maxiter=50, κ=0.5, δ=1e-5)
...
+function _optimal_transport_cost_plan(c, μ::DiscreteNonParametric, ν::DiscreteNonParametric)
+    cost = 0
```
devmotion, line 803: This initialization leads to a type instability: `cost` will be defined as an `Int` but below you will add floating point numbers to it. This should be avoided. One possibility is to initialize it with `cost = c(mu.support[1], nu.support[1])` for the first elements and remove this case from the loop below.

src/OptimalTransport.jl
```diff
@@ -672,4 +675,162 @@ function quadreg(mu, nu, C, ϵ; θ=0.1, tol=1e-5, maxiter=50, κ=0.5, δ=1e-5)
...
+    len_μ = length(μ.p)
+    len_ν = length(ν.p)
+    γ = zeros(len_μ, len_ν)
```
devmotion, line 806: Same here, this might not be of the correct types. Better to use
```suggestion
    γ = zeros(Base.promote_eltype(μ.p, ν.p), len_μ, len_ν)
```
In general, I think it might be better to output a sparse matrix - most values will be zero, so it seems more efficient to work with a sparse matrix.

test/runtests.jl
```diff
@@ -51,6 +52,46 @@ Random.seed!(100)
...
+    μ_n = μ_n / sum(μ_n)
+    ν_m = ν_m / sum(ν_m)
+
+    μ = DiscreteNonParametric(μ, μ_n)
+    ν = DiscreteNonParametric(ν, ν_m)
+
+    c(x, y) = (x - y)^2
+    # new version of StatsBase also has functoin pairwise,
+    # which conflicts with Distances.pairwise
+    C = Distances.pairwise(SqEuclidean(), μ.support, ν.support)
```
devmotion, line 81: No, there should not be any conflicts. They both just add methods to `StatsAPI.pairwise` and reexport it, it's exactly the same function.
```suggestion
    C = pairwise(SqEuclidean(), μ.support, ν.support)
```

test/runtests.jl
```diff
@@ -51,6 +52,46 @@ Random.seed!(100)
...
+
+    μ = rand(n)
+    ν = rand(m) .+ 0.5
+    μ_n = rand(n)
+    ν_m = rand(m)
+    μ_n = μ_n / sum(μ_n)
+    ν_m = ν_m / sum(ν_m)
+
+    μ = DiscreteNonParametric(μ, μ_n)
+    ν = DiscreteNonParametric(ν, ν_m)
```
devmotion, line 76: This is confusing for a reader and also for the compiler sometimes: one should not reassign different types to existing variable names. Better use `\mu_support` for the support.

test/runtests.jl
```diff
@@ -51,6 +52,46 @@ Random.seed!(100)
...
+
+    c(x, y) = (x - y)^2
```
devmotion, line 78: Just use `sqeuclidean`, no need to define `c` :slightly_smiling_face:

test/runtests.jl
```diff
@@ -51,6 +52,46 @@ Random.seed!(100)
...
+
+    # Discrete Case
+    n, m = 100, 150
+
+    μ = rand(n)
+    ν = rand(m) .+ 0.5
+    μ_n = rand(n)
+    ν_m = rand(m)
+    μ_n = μ_n / sum(μ_n)
+    ν_m = ν_m / sum(ν_m)
```
devmotion, line 73: This avoids some allocations caused by temporary arrays and makes it clearer what these arrays are:
```suggestion
    μ_probs = rand(n)
    ν_probs = rand(m)
    μ_probs ./= sum(μ_probs)
    ν_probs ./= sum(ν_probs)
```

test/runtests.jl
```diff
@@ -51,6 +52,46 @@ Random.seed!(100)
     @test cost2 ≈ cost
 end
 
+@testset "1D Optimal Transport for Convex Cost" begin
+    # Continuous Case
+    μ = Normal(0, 2)
+    ν = Normal(10, 2)
+    c(x, y) = abs(x - y)
+    γ = optimal_transport_plan(c, μ, ν)
```
devmotion, line 60:
```suggestion
    γ = optimal_transport_plan(euclidean, μ, ν)
```

test/runtests.jl
```diff
@@ -51,6 +52,46 @@ Random.seed!(100)
     @test cost2 ≈ cost
 end
 
+@testset "1D Optimal Transport for Convex Cost" begin
+    # Continuous Case
+    μ = Normal(0, 2)
+    ν = Normal(10, 2)
```
devmotion, line 58: Maybe use normal distributions with unequal standard deviations?

test/runtests.jl
```diff
@@ -51,6 +52,46 @@ Random.seed!(100)
...
+    ν = DiscreteNonParametric(ν, ν_m)
+
+    c(x, y) = (x - y)^2
+    # new version of StatsBase also has functoin pairwise,
+    # which conflicts with Distances.pairwise
+    C = Distances.pairwise(SqEuclidean(), μ.support, ν.support)
+
+    lp = Tulip.Optimizer()
+    cost_simplex = emd2(μ.p, ν.p, C, lp)
+    cost_1d = optimal_transport_cost(c, μ, ν)
```
devmotion, line 85:
```suggestion
    cost_1d = optimal_transport_cost(sqeuclidean, μ, ν)
```

test/runtests.jl
```diff
@@ -51,6 +52,46 @@ Random.seed!(100)
...
+
+    P = emd(μ.p, ν.p, C, lp)
+    γ = optimal_transport_plan(c, μ, ν)
```
devmotion, line 88:
```suggestion
    γ = optimal_transport_plan(sqeuclidean, μ, ν)
```

test/runtests.jl
```diff
@@ -51,6 +52,46 @@ Random.seed!(100)
...
+
+    @test cost_1d ≈ cost_simplex atol = 1e-5
+    @test dot(C, γ) ≈ dot(C, P) atol = 1e-5
```
devmotion, line 91: It seems a bit more straightforward to just check
```suggestion
    @test γ ≈ P atol = 1e-5
```

devmotion (changes requested):
I did another round of review :slightly_smiling_face: It seems you missed some of my previous comments but I added more comments and suggestions.

Project.toml
```diff
@@ -5,11 +5,15 @@ version = "0.2.3"
 
 [deps]
 Distances = "b4f34e82-e78d-54a5-968a-f98e89d6e8f7"
+Distributions = "31c24e10-a181-5473-b8eb-7969acd0382f"
```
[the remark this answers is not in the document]
davibarreira: How does one figure out what versions should be in the `[compat]`?

src/OptimalTransport.jl
davibarreira: Should I do the same for the `x = quantile (\mu,q)`?

Project.toml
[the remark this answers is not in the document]
devmotion: Either use the latest versions (since these are tested) or the oldest non-breaking release (e.g., 0.25 (=0.25.0) for Distributions instead of 0.25.1).

src/OptimalTransport.jl
devmotion: No, that's fine. It's not about avoiding all variable names but reducing unnecessary function definition since it creates overhead, both mentally and for the compiler.

src/OptimalTransport.jl
davibarreira: What would be more efficient than ` dot(StatsBase.pairwise(c, μ.support, ν.support), plan)`?

devmotion: Just accumulating the costs for the combinations with non-zero flow (which is easy if the transport plan is sparse).

docs/src/index.md
```diff
@@ -7,6 +7,13 @@ emd
 emd2
 ```
 
+### Exact optimal transport for 1D
+
+```@docs
+optimal_transport_cost
+optimal_transport_plan
```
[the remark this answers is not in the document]
zsteve: Hi @devmotion, 
Yes I agree that we can move away from mirroring the POT function naming scheme as currently. Although with some thought, I think earth mover's distance/EMD is clearer (at least for me) in that it refers specifically to the unregularised OT problem. So perhaps `emdcost` or `emdplan` would work? What do you think.

src/OptimalTransport.jl
davibarreira: I've tried this two options, and both performed worse than using the `pairwise`.
```
f(x) = sqeuclidean.(μ.support[x[1]],ν.support[x[2]])*γ[x]
```
And
```
function total(γ)
    im = findall(!iszero, γ)
    cost = γ[im[1]]*sqeuclidean(μ.support[im[1][1]],ν.support[im[1][2]])
    for i in im[2:end]
        cost+= γ[i]*sqeuclidean(μ.support[i[1]],ν.support[i[2]])
    end
    return cost
end
```
```

File: docs/src/index.md (hunk 44 above)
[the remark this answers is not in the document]
Comment (davibarreira) [reply docs/src/index.md:14]: My dissertation has been a literature review on Optimal Transport for Machine Learning. In this kind of literature, the term Earth-Movers Distance is not very usual (it's mostly used in the earlier papers), and it usually refers to the case where the cost is |x-y|, hence the name. So for me, I find it really confusing, because when I use the `emd` function, I tend to think that I'm only dealing with the distance as cost.

I find `otcost` and `otplan` to be the most direct and concise. I guess @devmotion  will be the decisive vote.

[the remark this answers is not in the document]
Comment (devmotion) [reply docs/src/index.md:14]: I think I tend to `otcost` and `otplan` since it seems reasonably descriptive and still concise. For me, the terms optimal transport cost and plan already imply that we operate in the unregularized setting (similar to other areas in e.g. ML and statistics where maximum likelihood estimation is unregularized by default and one only specifies explicitly if it is regularized). In my opinion, the term is also better than "earth mover's distance" since the latter seems to be a more recently introduced name in the field of computer vision (see e.g. the introduction in Computational Optimal Transport and the [Wikipedia page to Earth mover's distance](https://en.wikipedia.org/wiki/Earth_mover%27s_distance#History)).

That being said, we can always revisit later and rename the function if it turns out that it was a bad choice.

File: src/OptimalTransport.jl [hunk 45]
```diff
@@ -672,4 +675,162 @@ function quadreg(mu, nu, C, ϵ; θ=0.1, tol=1e-5, maxiter=50, κ=0.5, δ=1e-5)
...
+``h`` is a convex function.
+
+Parameters:\n
+c: The cost function, which should be of the form ``c(x,y) = h(abs(x-y))``
+where `h` is a convex function.
+mu: 1-D distribution (e.g. `mu = Distributions.Normal(0,1)`)\n
+nu: 1-D distribution (e.g. `nu = Distributions.Normal(0,1)`)\n
+plan: Optional parameter in case the optimal transport plan is already known. Providing
+it can speed up calculations.
+"""
```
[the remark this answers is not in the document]
davibarreira: I was avoiding the four spaces in the first line, because this is causing an error with Jupyter when rendering the docs. For some reason, if you add spaces in the first line, the math equations are not rendered when you visualize the documentation with jupyter.
But I guess this is an issue more with jupyter than with the docstring, so I'll use the standard.

docs/src/index.md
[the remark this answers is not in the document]
zsteve: Ok yes, I agree. I think we should separate the words with an underscore `ot_cost` and `ot_plan`, since `sinkhorncost` and `sinkhornplan` are a bit long.

src/OptimalTransport.jl
```diff
@@ -672,4 +675,156 @@ function quadreg(mu, nu, C, ϵ; θ=0.1, tol=1e-5, maxiter=50, κ=0.5, δ=1e-5)
...
+    _ot_cost_plan(c, μ::DiscreteNonParametric, ν::DiscreteNonParametric; get=:plan)
+
+Compute the optimal transport cost and plan for
+μ an ν where both are 1-Dimensional discrete distributions with finite support
+and the cost function is of the form ``c(x,y) = h(|x-y|)`` such that
+``h`` is a convex function.
+
+Return the plan or the cost depending if `get=:plan` or `get=:cost`, respectively.
+"""
+function _ot_cost_plan(c, μ::DiscreteNonParametric, ν::DiscreteNonParametric; get=:plan)
```
devmotion, line 783: I don't think that's a good approach, the function is type unstable and we can just hope that constant propagation kicks in. I had in mind the following: https://gist.github.com/devmotion/6b987174bce2bedf18b074358be80198 Maybe just leave out the `DiscreteNonParametric` case and I open a separate PR after basic support for Distributions is merged?

devmotion (commented):
Looks mostly OK, can you change `otcost` and `otplan` to `ot_cost` and `ot_plan` as suggested by @zsteve? And maybe just leave out the DiscreteNonParametric case for now?

src/OptimalTransport.jl
davibarreira: Sure, I'll just submit the other part.
If I may, I'd like to understand some of the choices you made regarding the code in the gist. I see that you kept the `_ot_cost_plan` function with the `get`. Is this just for show or you intend to use it?
Secondly, why do you create this `Base.iterate` function.
I do not quite understand the issue with the function. I mean, what do you mean by "we can just hope that constant propagation kicks in"? How could this code go wrong?
Thanks

devmotion: The `_ot_cost_plan` I just included as a comparison and to check that the other implementation gives the same results.

The `Base.iterate` function (and also `Base.IteratorSize` and `Base.IteratorEltype`) belong to the iteration interface in Julia: https://docs.julialang.org/en/v1/base/collections/#lib-collections-iteration You have to define them if you implement an iterator. The custom iterator in the gist will just iteratively return tuples `(i, j, w)` where `i` and `j` are the indices of the source and target and `w` is the flow between them. This can be used to compute the cost and the plan completely independently (one could also compute both at the same time).

It is problematic if the return type of a function can't be computed from the types of the inputs. This is so-called type instability. Since the compiler does not know the type of the result, it can't create optimized compiled code and instead has to fall back on dynamically operating with the result at runtime (which typically is significantly slower). In this function the return type depends on the value of `get` and therefore it can't be inferred from the input types.

If you hardcode literals such as `1` or `:cost` then sometimes the compiler is smart enough and is able to infer the return type nevertheless. This is so-called constant propagation. Even though constant propagation improved a lot in recent Julia versions there are no hard guarantees and it might work or not.

BTW a common way to deal with type-unstable functions (sometimes it is not possible to avoid type instabilities) are so-called function barriers: if the result whose type can't be inferred is passed on to another function then the compiler still optimizes the called function for the type that it was called with, so the type instability does not impact the performance of subsequent computations. However, to achieve this you have to call another function and therefore it's called function barrier.

davibarreira: Sorry for taking so long to submit the PR. As asked, I removed the `DiscreteNonParametric` implementation. I've left the tests commented out, so you can just uncomment when submitting your PR. I'm interested to see how you'll implement this. I've read your gist, but didn't quite understand all the code, and the implementation from `POT.py` returns both the cost and the transport plan.

davibarreira:
I think one of the tests might be using a too restrictive tolerance for the error. The test failing for the macOS is not even the 1D case.

davibarreira:
@devmotion , can you merge this PR?

devmotion:
It seems you reintroduced the notebook :stuck_out_tongue: Is it OK, if I clean it up a bit and then if you're OK with the changes, we can merge it?

davibarreira:
> It seems you reintroduced the notebook  Is it OK, if I clean it up a bit and then if you're OK with the changes, we can merge it?

I did?! My bad, let me remove it.

davibarreira:
> It seems you reintroduced the notebook  Is it OK, if I clean it up a bit and then if you're OK with the changes, we can merge it?

Sure, you can make any changes you like.

devmotion:
I reverted the changes where I tried to debug the MacOS test failures. It's better to not include any such fixes in the PR, hopefully https://github.com/JuliaOptimalTransport/OptimalTransport.jl/pull/79 and similar PRs can fix the problem.

Otherwise, I cleaned the code and the docstrings a bit and added more tests. @davibarreira do you have any additional comments or suggestions?

davibarreira:
@devmotion , is there something still pending on this PR?

devmotion (approved):

devmotion:
My idea was to fix the test failures first but I don't want to rush the other PRs and this seems good to go, so I'll merge even if some tests might fail on master.

PR #46, opened 2021-05-15
OT for Multivariate Normal

davibarreira:
For the 2-Wasserstein distance, OT for two the Multivariate Normal has closed form solution. Hence, I've implemented 3 functions:
* `w2mvncost`
* `w2mvnplan`
* `w2mvncostplan`

I've already updated the documentation and added tests for each of the functions. I've decided to write the `w2mvncostplan`, because it reduces the number of calculations if you instead calculate the cost and the plan in two different functions.


devmotion:
Again, I don't think it is a good idea to introduce special functions for this case. We should just always use `optimal_transport_plan` and `optimal_transport_cost` (or `otplan` and `otcost` but personally I would prefer the more descriptive versions, as usually more descriptive names are preferred in Julia) and use multiple dispatch. There should not be a need for different functions.

src/OptimalTransport.jl
```diff
@@ -112,6 +116,109 @@ function emd2(μ, ν, C, optimizer)
...
+  \\Sigma_\\mu^{1/2} \\Sigma_\\nu
+  \\Sigma_\\mu^{1/2}
+  \\right)^{1/2}
+  \\Sigma_\\mu^{-1/2}.
+```
+"""
+function w2mvncost(X::MvNormal, Y::MvNormal)
+    XΣsqrt = X.Σ^(1/2)
+    w2     = sqrt(norm(X.μ - Y.μ)^2 + tr(X.Σ + Y.Σ - 2*(XΣsqrt*Y.Σ*XΣsqrt)^(1/2)))
+    return w2
```
devmotion, line 151: I think it is better to use dedicated functions for the Bures metric that allow optimizations based on the type of the covariance matrix. For many cases the computations here will be quite inefficient.

You can have a look at how I implemented it in https://github.com/devmotion/CalibrationErrorsDistributions.jl/blob/85d8ca7a71b319b31d64b520aff344fd6c6758e6/src/distances/wasserstein.jl#L28 and https://github.com/devmotion/CalibrationErrorsDistributions.jl/blob/main/src/distances/bures.jl.

davibarreira:
I agree that we can put it all together in one function. But since @zsteve has not yet given his input, I wrote the different functions. If we decide to put it all under one function, then I'll alter everything and send a pull request with both 1D and the Normal case under the same function. The only complication I see with this case of the Multivariate Normals is that it only works if the cost is the squared. How do we enforce this with multiple dispatch? Also, I think we should emphasize  somewhere that these cases (1D and Normal 2-W) are more efficient and have closed form. Perhaps we can point it out in the docs.

src/OptimalTransport.jl
davibarreira: Great! I was wondering if there was a more efficient way of doing this, but I did not find this implementation on POT.py.
I'll take a look at your code and modify.

devmotion:
> The only complication I see with this case of the Multivariate Normals is that it only works if the cost is the squared. How do we enforce this with multiple dispatch? Also, I think we should emphasize somewhere that these cases (1D and Normal 2-W) are more efficient and have closed form. Perhaps we can point it out in the docs.

If we add a docstring then it will show up in the REPL if someone looks up `? optimal_transport_map` etc. And, of course, it should be included in the documentation. For this special case we should just dispatch on `optimal_transport_plan(::SqEuclidean, ::MvNormal, ::MvNormal)` etc.

devmotion:
Also related here is my high-level suggestion in https://github.com/zsteve/OptimalTransport.jl/pull/45#issuecomment-841677860.

src/OptimalTransport.jl
davibarreira: I see that use used `PDMats.jl` in the calculations for the Bures metric. So you suggests we also add that as a dependency?

devmotion: Distributions already depends on it :shrug:

davibarreira: Ok then. I saw that you've implemented various cases already, so I'll just reuse your code on for the Bures calculation.

davibarreira:
What is missing for this PR to be accepted? Any other changes I should make?

devmotion:
Can you remove the changes from the other PR? It's difficult to review since the PR also contains unrelated changes.

davibarreira:
Sure, I'll close this PR and submit a new one.

PR #48, opened 2021-05-17
Use BlueStyle

devmotion:
Now that we see more activity and contributions in this repository, I think it would be good to adopt some style guide. This helps to keep the code base consistent and, in my opinion, increases readability and reduces mental overhead.

I just chose [BlueStyle](https://github.com/invenia/BlueStyle) here since it is used in most projects I contribute to. An other popular style for Julia is YAS. I would be fine with it as well, I think a main criterion should just be that is supported by [JuliaFormatter](https://github.com/domluna/JuliaFormatter.jl) since this makes it much easier to format code correctly.

Speaking about convenience, I also added a configuration for JuliaFormatter such that one can just run `format(path_to_repo)` to re-format the codebase correctly (actually, that's what I did when putting together the PR). Moreover, I added a Github action that will automatically check and report if PRs are formatted correctly. It will even post suggestions with format fixes (if required) which is even more convenient.

What do you think, @zsteve @davibarreira?

### DIFF
--- a/.JuliaFormatter.toml
+++ b/.JuliaFormatter.toml
@@ -1,0 +1,1 @@
+style="blue"
--- a/.github/workflows/Format.yml
+++ b/.github/workflows/Format.yml
@@ -1,0 +1,24 @@
+name: Format
+
+on:
+  pull_request:
+
+jobs:
+  format:
+    runs-on: ubuntu-latest
+    steps:
+      - uses: actions/checkout@v2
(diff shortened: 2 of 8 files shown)


davibarreira:
Sounds great. For me Blues is fine.

zsteve:
Yes, this is excellent! Thanks for taking the initiative for this!

PR #50, opened 2021-05-17
Add pre-computed map

devmotion:
This PR adds a `map` keyword argument that allows to specify pre-computed optimal transport maps when computing the optimal transport cost.

I suggested this approach in https://github.com/zsteve/OptimalTransport.jl/pull/45 to avoid having to solve the optimal transport problem twice. The approach is used, e.g., also in Statistics for [std](https://docs.julialang.org/en/v1/stdlib/Statistics/#Statistics.std) and [var](https://docs.julialang.org/en/v1/stdlib/Statistics/#Statistics.var).

I am not sure if it would be better to use the name `plan` for the keyword argument. It's fine with me either way, I went for `map` since it seems other docstrings in this package mostly use the term `map`. In any case, I think we should try to be somewhat consistent throughout the package.

### DIFF
--- a/Project.toml
+++ b/Project.toml
@@ -1,7 +1,7 @@
 name = "OptimalTransport"
 uuid = "7e02d93a-ae51-4f58-b602-d97af76e3b33"
 authors = ["zsteve <stephenz@student.unimelb.edu.au>"]
-version = "0.2.1"
+version = "0.2.2"
 
 [deps]
 Distances = "b4f34e82-e78d-54a5-968a-f98e89d6e8f7"
--- a/src/OptimalTransport.jl
+++ b/src/OptimalTransport.jl
@@ -92,7 +92,7 @@ function emd(μ, ν, C, model::MOI.ModelLike)
 end
 
 """
-    emd2(μ, ν, C, optimizer)
+    emd2(μ, ν, C, optimizer; map=nothing)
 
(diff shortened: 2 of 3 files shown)


zsteve:
Yep, using the map keyword is fine. I'll update the docstrings to use map consistently.

davibarreira:
> This PR adds a `map` keyword argument that allows to specify pre-computed optimal transport maps when computing the optimal transport cost.
> 
> I suggested this approach in #45 to avoid having to solve the optimal transport problem twice. The approach is used, e.g., also in Statistics for [std](https://docs.julialang.org/en/v1/stdlib/Statistics/#Statistics.std) and [var](https://docs.julialang.org/en/v1/stdlib/Statistics/#Statistics.var).
> 
> I am not sure if it would be better to use the name `plan` for the keyword argument. It's fine with me either way, I went for `map` since it seems other docstrings in this package mostly use the term `map`. In any case, I think we should try to be somewhat consistent throughout the package.

I missed this thread. But I think we should use the word  `plan` instead of `map`, since in a `plan` is always a `map`, but the `map` is not always a `plan`, since the `map` is deterministic.

devmotion:
That's a good point :+1: I use the terms interchangeably but I guess it is more common to only denote Monge maps as maps.

Issue #42, opened 2021-04-29
Separate Python even more?


devmotion:
I am implementing a Python interface for one of my packages with pyjulia. The package depends on OptimalTransport but since POT was quite annoying to install (there's a bug in pip and therefore one always has to install cython and numpy first, otherwise installation will fail) I completely switched to OptimalTransport + Tulip, eliminating all Python dependencies. However, pyjulia requires PyCall and installs and loads it automatically, and therefore automatically also the POT submodule in OptimalTransport is loaded even though I don't want to use it. And now Python users either have to install POT which is quite annoying and also problematic for CI (and it's not even used at all!), or they'll get an error message when OptimalTransport tries to load it in https://github.com/zsteve/OptimalTransport.jl/blob/9ad8f733770b7cd446c84ebfc90e2a558d6d120b/src/pot.jl#L8.

Maybe the Python interface to POT could be moved to a completely separate package or one could require users to load, and possibly install, it explicitly by making the `__init__` function a proper function that could be called e.g. `load_POT!`.

devmotion:
I fixed the issue for now by removing the dependency on OptimalTransport and reimplementing the LP formulation in my package. I'll open a PR with the version in my package since it seems to be simpler, more efficient, and requires fewer dependencies.

devmotion:
I extracted the POT interface into a separate package: https://github.com/devmotion/POT.jl

zsteve:
Hi @devmotion ,
Thanks for the proposal and for contributing the separate package! This sounds fine, and I agree that moving forward separating out the POT component will make things easier in terms of dependencies and also in general.
I will reference the POT.jl package in the README.

PR #55, opened 2021-05-18
Test quadratic optimal transport

zsteve:
I've added a test for `quadreg()`. As with the other functions, I compare against the output of POT. The relevant function here is `smooth_ot_dual` from POT, and I've needed to implement another wrapper for it. (I agree that eventually it would be nice to strip the python dependencies, but we can talk about how to do testing in #42).

### DIFF
--- a/Project.toml
+++ b/Project.toml
@@ -8,9 +8,7 @@ Distances = "b4f34e82-e78d-54a5-968a-f98e89d6e8f7"
 IterativeSolvers = "42fd0dbc-a981-5370-80f2-aaf504508153"
 LinearAlgebra = "37e2e46d-f89d-539d-b4ee-838fcccc9c8e"
 MathOptInterface = "b8f27783-ece8-5eb3-8dc8-9495eed66fee"
-PyPlot = "d330b81b-6aea-500a-939a-2ce795aea3ee"
 Requires = "ae029012-a4dd-5104-9daa-d747884805df"
-Revise = "295af30f-e4ad-537b-8983-00126c2a3abe"
 SparseArrays = "2f01184e-e22b-5df5-ae63-d93ebab69eaf"
 
 [compat]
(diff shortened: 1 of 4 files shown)


Project.toml
```diff
@@ -8,7 +8,9 @@ Distances = "b4f34e82-e78d-54a5-968a-f98e89d6e8f7"
 IterativeSolvers = "42fd0dbc-a981-5370-80f2-aaf504508153"
 LinearAlgebra = "37e2e46d-f89d-539d-b4ee-838fcccc9c8e"
 MathOptInterface = "b8f27783-ece8-5eb3-8dc8-9495eed66fee"
+PyPlot = "d330b81b-6aea-500a-939a-2ce795aea3ee"
 Requires = "ae029012-a4dd-5104-9daa-d747884805df"
+Revise = "295af30f-e4ad-537b-8983-00126c2a3abe"
```
devmotion, line 13: It seems you committed some local dependencies:
```suggestion
Requires = "ae029012-a4dd-5104-9daa-d747884805df"
```

src/pot.jl
```diff
@@ -154,4 +154,19 @@ function sinkhorn_unbalanced2(
     )[1]
 end
 
+function smooth_ot_dual(
+    mu, nu, C, eps; reg_type="l2", method="L-BFGS-B", tol=1e-9, max_iter=500, verbose=false
+)
+    return pot.smooth.smooth_ot_dual(
+        nu,
+        mu,
```
devmotion, line 162: Just one to mention that I changed the Python wrappers in POT.jl such that it is not necessary to pass the arguments in the different order and all POT keyword arguments are supported and set to their Python defaults.

devmotion (commented):
Always good to add more tests :+1: 

I wonder though if the the new POT function should only be defined (or just be called directly) in the tests. I initiated the registration process of POT.jl some days ago, and just wait for feedback regarding the name (https://github.com/JuliaRegistries/General/pull/36503), and then it should be available official in the package manager. So it should not take long until the Python submodule can be removed completely from OptimalTransport. Moreover, the implementations in POT.jl are a bit more general and fix some problems with the current wrapper, so there is an additional incentive to retire the submodule :slightly_smiling_face:

src/pot.jl
zsteve: Got it, sounds good. Once POT.jl is available from Pkg I can create a pull request with updated tests.

PR #56, opened 2021-05-18
Tests for sinkhorn_barycenter

zsteve:
Trying to get the code coverage up to a healthy percentage :)
I've noticed a possible bug with PythonOT where it appears to ignore the `weights` parameter supplied, and automatically take uniform weights. So I've just tested this case for now. With further investigation it would be helpful to use non-uniform weights.

### DIFF
--- a/Project.toml
+++ b/Project.toml
@@ -8,9 +8,7 @@ Distances = "b4f34e82-e78d-54a5-968a-f98e89d6e8f7"
 IterativeSolvers = "42fd0dbc-a981-5370-80f2-aaf504508153"
 LinearAlgebra = "37e2e46d-f89d-539d-b4ee-838fcccc9c8e"
 MathOptInterface = "b8f27783-ece8-5eb3-8dc8-9495eed66fee"
-PyPlot = "d330b81b-6aea-500a-939a-2ce795aea3ee"
 Requires = "ae029012-a4dd-5104-9daa-d747884805df"
-Revise = "295af30f-e4ad-537b-8983-00126c2a3abe"
 SparseArrays = "2f01184e-e22b-5df5-ae63-d93ebab69eaf"
 
 [compat]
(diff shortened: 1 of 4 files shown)


Project.toml
```diff
@@ -8,7 +8,9 @@ Distances = "b4f34e82-e78d-54a5-968a-f98e89d6e8f7"
 IterativeSolvers = "42fd0dbc-a981-5370-80f2-aaf504508153"
 LinearAlgebra = "37e2e46d-f89d-539d-b4ee-838fcccc9c8e"
 MathOptInterface = "b8f27783-ece8-5eb3-8dc8-9495eed66fee"
+PyPlot = "d330b81b-6aea-500a-939a-2ce795aea3ee"
 Requires = "ae029012-a4dd-5104-9daa-d747884805df"
+Revise = "295af30f-e4ad-537b-8983-00126c2a3abe"
```
devmotion, line 13: Again better remove these :wink: 

```suggestion
Requires = "ae029012-a4dd-5104-9daa-d747884805df"
```

src/OptimalTransport.jl
```diff
@@ -115,8 +115,9 @@ function emd2(μ, ν, C, optimizer; map=nothing)
         # check dimensions
         size(C) == (length(μ), length(ν)) ||
             error("cost matrix `C` must be of size `(length(μ), length(ν))`")
-        size(map) == size(C) ||
-            error("optimal transport map `map` and cost matrix `C` must be of the same size")
+        size(map) == size(C) || error(
+            "optimal transport map `map` and cost matrix `C` must be of the same size",
+        )
```
devmotion, line 120: Weren't these changes also part of your last PR? :thinking:

Project.toml
zsteve: Sorry! You're right. I should check the Project.toml before I pull request. I think it happens when I activate the current environment and then add packages.

PR #58, opened 2021-05-18
New logo

davibarreira:
Updated the colors as requested.


zsteve:
Looks like the README is missing the graphic for me. Does it work to use the .svg in the readme?

devmotion:
What do you think about having a proper regular level 1 heading and just the logo on the right? I think it looks quite neat in https://github.com/devmotion/PalmerPenguins.jl (I copied it from https://github.com/allisonhorst/palmerpenguins), and the heading is still present even if the image isn't/can't be loaded.

davibarreira:
Yeah, it works with "svg". Maybe it just doesn't work on the Preview. I'll try the layout of https://github.com/devmotion/PalmerPenguins.jl .

devmotion:
> Yeah, it works with "svg". Maybe it just doesn't work on the Preview.

I assume the problem is that you deleted the png file (which is fine with me, SVG should be of higher quality and is not a binary file).

PR #64, opened 2021-05-23
Use PythonOT

### LINKED ISSUE #42: Separate Python even more?

devmotion:
This PR removes the Python interface from the package and instead uses PythonOT in the tests. I couldn't find a way to remove the submodule in a non-breaking way since we don't want to make PyCall or PythonOT a non-optional dependency now but optional fallbacks would require that users have already installed PythonOT which defeats the purpose.

Fixes https://github.com/JuliaOptimalTransport/OptimalTransport.jl/issues/42.


devmotion:
It took a while but now all tests pass: I simplified the CI tests and removed the manual installation of Python and POT (handled by Conda.jl now) but this caused test failures. I fixed these numerical issues by increasing the maximum number of iterations and adjusting tolerances in POT. I also used the opportunity to add tests for Julia 1.0 which uncovered some bugs.

devmotion:
I separated the changes related to Julia 1.0, the remaining changes are all related to PythonOT and testing with Conda.jl.

zsteve:
> It took a while but now all tests pass: I simplified the CI tests and removed the manual installation of Python and POT (handled by Conda.jl now) but this caused test failures. I fixed these numerical issues by increasing the maximum number of iterations and adjusting tolerances in POT. I also used the opportunity to add tests for Julia 1.0 which uncovered some bugs.

Did these numerical issues arise from changing the route of installation of POT? Strange

zsteve (approved):

devmotion:
> Did these numerical issues arise from changing the route of installation of POT? Strange

I think it might be also due to different Python versions - previously we installed a specific version explicitly whereas now just the default miniconda Python version of Conda.jl is used.

Issue #63, opened 2021-05-22
Deprecating and renaming additional functions


zsteve:
Similar to #45, a lot of functions currently are either named mirroring Python OT's function naming choices, or just ad-hoc as they were added originally. (I agree with `ot_cost` and `ot_plan` for unregularised OT as per #45, for the record).

- Along those lines, for the `sinkhorn` routines we should use `sinkhorn_cost` and `sinkhorn_plan`, etc. 
- What to rename `quadreg` to? This solves OT regularised with an L2 cost on the transport plan. Currently it's solved using one particular algorithm, but there are other approaches, so I'd prefer not to refer to any particular algorithm in the function name.
One possibility is to have a general `ot_reg_cost` and `ot_reg_plan` wrapper for generalised regularisations. Then the user can specify the regularisation functional and algorithm upon calling. So something like 
`ot_reg_cost(mu, nu, C, 0.05; reg_func = "L2", method = "lorenz")`

devmotion:
I agree, I think a more concise and easier accessible API would be nice. E.g., I guess (actually similar to `sinkhorn` and `sinkhorn2` in POT) it would be good if there is only one `sinkhorn_cost` and `sinkhorn_plan` function instead of one for the standard algorithm, the stabilized version and the stabilized version with epsilon scaling.

I am not sure if the API for all regularized optimal transport problems should be unified, maybe it could become too confusing with all the keyword arguments - not all algorithms would support them and their meaning might change depending on the algorithm. Maybe instead of `ot_reg_cost(mu, nu, C, 0.05; reg_func = "L2", method = "lorenz")` one could use `ot_regularized_cost(mu, nu, C, 0.05; regularization=:L2, method=Lorenz(algorithm_specific_options...), general_options...)`? Here, `general_options` could contain e.g. `plan`. Additionally, maybe one could even make the regularization parameter a mandatory keyword argument, i.e., one without default value, to keep the API closer to the one for exact OT problems.

A different idea would be to adopt the CommonSolve interface used by SciML and also e.g. Roots. Mainly, it would be something along
```julia
abstract type AbstractOTProblem end

struct OTProblem <: AbstractOTProblem
    cost::C
	mu::M
    nu::N
end

struct RegularizedOTProblem <: AbstractOTProblem
    cost::C
	mu::M
    nu::N
    regularization::R
	regularization_parameter::P
end

struct L2Regularization end
struct EntropyRegularization end

[...]

PR #66, opened 2021-05-23
General regularisation

zsteve:
Instead of using the `quadreg` function call for the quadratic regularisation, this PR provides a general set of functions `ot_reg_plan` and `ot_reg_cost` for computing OT with 'general' regularisations. Now `quadreg` is wrapped as `reg_func = "L2"`, and `method = "lorenz"` selects the current algorithm. However, there are other approaches out there (e.g. L-BFGS-B) for the same OT problem that could be implemented using the same interface.

The idea is that eventually more regularisations can be added (see e.g. https://arxiv.org/abs/1710.06276), and there should be a unified interface for this.


src/OptimalTransport.jl
```diff
@@ -534,6 +534,53 @@ function sinkhorn_barycenter(
...
+
+Supported solution methods are:
+- L2: `method = "lorenz"` for the semi-smooth Newton method of Lorenz et al. 
+
+References
+
+Lorenz, D.A., Manns, P. and Meyer, C., 2019. Quadratically regularized optimal transport. Applied Mathematics & Optimization, pp.1-31.
+"""
+function ot_reg_plan(mu, nu, C, eps; reg_func="L2", method="lorenz", kwargs...)
+    if (reg_func == "L2") && (method == "lorenz")
```
devmotion, line 558: This approach is problematic in my opinion: it is not possible to add support for other methods or regularization types for users or downstream packages, one always has to modify this function here.

I guess this could be avoided with the suggestion in https://github.com/JuliaOptimalTransport/OptimalTransport.jl/issues/63#issuecomment-846618049 - everyone could just add other regularizations and/or algorithms.

davibarreira:
Can this PR be closed, @zsteve ?

PR #67, opened 2021-05-23
Enable batch computation of OT via Sinkhorn

zsteve:
I've modified the standard Sinkhorn solver `sinkhorn_gibbs` to allow batch computations for multiple histograms, as well as `sinkhorn_barycenter` to handle the special case where all cost matrices are identical. The modification is to go from matrix-vector to matrix-matrix multiplication. 

This does not break existing uses of Sinkhorn, but the API for `sinkhorn_barycenter` has changed slightly (old implementation took histograms as rows, now taken to be columns).


src/OptimalTransport.jl
```diff
@@ -192,7 +199,11 @@ function sinkhorn(mu, nu, C, eps; kwargs...)
     # compute dual potentials
     u, v = sinkhorn_gibbs(mu, nu, K; kwargs...)
 
-    return Diagonal(u) * K * Diagonal(v)
+    if size(mu, 2) == 1
+        return Diagonal(reshape(u, :)) * K * Diagonal(reshape(v, :))
+    else
+        return [Diagonal(u[:, i]) * K * Diagonal(v[:, i]) for i in 1:size(mu, 2)]
+    end
```
devmotion, line 206: This creates a type instability: it is not possible to infer from the type of the inputs the type of the output of the function.

zsteve: Thanks for pointing that out. One alternative would be to use instead `mu isa Matrix` to resolve this issue.

src/OptimalTransport.jl
```diff
@@ -192,7 +199,11 @@ function sinkhorn(mu, nu, C, eps; kwargs...)
     # compute dual potentials
     u, v = sinkhorn_gibbs(mu, nu, K; kwargs...)
 
-    return Diagonal(u) * K * Diagonal(v)
+    if mu isa Vector
+        return Diagonal(reshape(u, :)) * K * Diagonal(reshape(v, :))
```
devmotion: A bit simpler:
```suggestion
        return Diagonal(vec(u)) * K * Diagonal(vec(v))
```

src/OptimalTransport.jl
zsteve: Hang on, I'm not sure that works :/ Let me take another look at this

src/OptimalTransport.jl
```diff
@@ -192,7 +199,11 @@ function sinkhorn(mu, nu, C, eps; kwargs...)
     # compute dual potentials
     u, v = sinkhorn_gibbs(mu, nu, K; kwargs...)
 
-    return Diagonal(u) * K * Diagonal(v)
+    if mu isa Vector
```
devmotion: It seems a bit too restrictive, maybe `AbstractVector` would be better
```suggestion
    if mu isa AbstractVector
```
or
```suggestion
    if ndims(mu) == 1
```
(this can be inferred from the type of `AbstractArray`).

src/OptimalTransport.jl
```diff
@@ -192,7 +199,11 @@ function sinkhorn(mu, nu, C, eps; kwargs...)
...
+        return Diagonal(reshape(u, :)) * K * Diagonal(reshape(v, :))
+    else
+        return [Diagonal(u[:, i]) * K * Diagonal(v[:, i]) for i in 1:size(mu, 2)]
```
devmotion: It seems this assumes that `u` and `v` contain the same number of histograms. I don't think this is checked anywhere?

src/OptimalTransport.jl
```diff
@@ -125,14 +125,17 @@ The Gibbs kernel `K` is given by `K = exp.(- C / eps)` where `C` is the cost mat
...
 """
 function sinkhorn_gibbs(mu, nu, K; tol=1e-9, check_marginal_step=10, maxiter=1000)
-    if !(sum(mu) ≈ sum(nu))
+    if !(sum(mu; dims=1) ≈ sum(nu; dims=1))
         throw(ArgumentError("Error: mu and nu must lie in the simplex"))
     end
-
     # initial iteration
-    temp_v = vec(sum(K; dims=2))
+    temp_v = repeat(vec(sum(K; dims=2)); outer=(1, size(mu, 2)))
```
devmotion, line 138: I assume this can be simplified by changing the algorithm to
```julia
u = mu ./ vec(sum(K; dims=2))
temp_u = K' * u
v = nu ./ temp_u
temp_v = K * v
```
and adjusting the `for` loop accordingly (i.e. move the `mul!(temp_v, K, v)` to the end of the computations). Then `temp_v` (and the other variable) should be of the correct shape automatically.

src/OptimalTransport.jl
```diff
@@ -184,15 +187,25 @@ Return optimal transport coupling `γ` of the same dimensions as `C` which solve
 
 where ``H`` is the entropic regulariser, ``H(\\gamma) = -\\sum_{i, j} \\gamma_{ij} \\log(\\gamma_{ij})``.
 
+Note that for a common cost `C`, multiple histograms may be provided for a batch computation by passing `mu` and `nu`
+as matrices whose columns `mu[:, i]` and `nu[:, i]` correspond to pairs of histograms. 
+The output in this case is a vector `γ` of coupling matrices such that `γ[i]` is a coupling of `mu[:, i]` and `nu[:, i]`.
+
 """
-function sinkhorn(mu, nu, C, eps; kwargs...)
+function sinkhorn(mu::Vector, nu::Vector, C, eps; kwargs...)
```
devmotion, line 195: I don't think it's a good idea to add these type annotations, they are too restrictive. I think it should work if you change `sinkhorn_gibbs` as suggested, then you can change computations depending on the resulting types of `u` and `v`.

zsteve: Ok, I've adopted the suggested changes (see new commits). By 'restrictive' do you mean some of the code is duplicated? I wasn't sure whether checking the type of the output as I commented originally would resolve the problem of type stability (I guess it does then!)

src/OptimalTransport.jl
```diff
@@ -184,15 +194,23 @@ Return optimal transport coupling `γ` of the same dimensions as `C` which solve
...
     u, v = sinkhorn_gibbs(mu, nu, K; kwargs...)
-
-    return Diagonal(u) * K * Diagonal(v)
+    if (u isa Vector) && (v isa Vector)
+        return Diagonal(vec(u)) * K * Diagonal(vec(v))
+    elseif (u isa Matrix) && (v isa Matrix)
+        return [Diagonal(u[:, i]) * K * Diagonal(v[:, i]) for i in 1:size(mu, 2)]
+    else
+        throw(ArgumentError("Error: sinkhorn_gibbs returned inconsistent outputs"))
+    end
```
devmotion, line 213: I have a different suggestion that would also add support for arbitrary arrays of histograms that can be broadcasted together (e.g., one source histogram and multiple target histograms and vice versa). The suggestion would be to write
```suggestion
    plan = K .* add_singleton(u, Val(2)) .* add_singleton(v, Val(1))
    return plan
```
and to define
```suggestion
function add_singleton(x::AbstractArray, ::Val{dim}) where {dim}
    shape = ntuple(ndims(x) + 1) do i
        return i < dim ? size(x, i) : (i > dim ? size(x, i - 1) : 1)
    end
    return reshape(x, shape)
end
```
Then with one source and one target histogram of length `m` and `n` a plan of size `(m, n)` would be returned. If there a `k` source histograms (size `(m, k)`), `k` plans are returned as array of size `(m, n, k)`. If there are `l` target histograms (size `(n, l)`), `l` plans are returned as array of size `(m, n, l)`. If there are `k` source and target histograms (size `(m, k)` and `(n, k)`), `k` pairwise plans are returned as array of size `(m, n, k)`. And if there are `k` source histograms (size `(m, k)`) and `l` target histograms (size `(n, 1, l)`), then `k x l` plans are returned as array of size `(m, n, k, l)`.

src/OptimalTransport.jl
devmotion: No, restrictive referred to the fact that one might want to use different vector types and not only `Vector` (e.g. static `SVector`s, `FillArray`s, etc.). If you dispatch on `Vector` or include checks for `Vector` (same for `Matrix`), all these other types are not covered and will either error or produce incorrect results.

zsteve: Got it!

src/OptimalTransport.jl
zsteve: Ok I'll give this a try. It certainly seems more natural than returning either a Matrix or Vector{Matrix}. 

One issue though, I don't think this will work in the setting where there are a different number of source and target histograms, since AFAIK the dual variables can't be mixed and matched between different OT instances. I.e. to compute OT from `p` to `[q1, ..., qn]` still requires `2n` dual variables.

zsteve: Ok I've adopted this change. `sinkhorn2` now returns in all cases an `Array` (never a scalar), for the sake of type stability. The output is the same as that of `PythonOT.sinkhorn2`.

src/OptimalTransport.jl
```diff
@@ -211,21 +230,26 @@ where ``H`` is the entropic regulariser, ``H(\\gamma) = -\\sum_{i, j} \\gamma_{i
...
     else
         # check dimensions
-        size(C) == (length(μ), length(ν)) ||
-            error("cost matrix `C` must be of size `(length(μ), length(ν))`")
+        size(C) == (size(μ, 1), size(ν, 1)) || error(
+            "cost matrix `C` must be of size `(size(μ, dims = 1), size(ν, dims = 1))`",
+        )
         size(plan) == size(C) || error(
             "optimal transport plan `plan` and cost matrix `C` must be of the same size",
         )
```
devmotion, line 249: This should not be enforced anymore, better
```suggestion
        (size(plan, 1), size(plan, 2)) == size(C) || error(
            "optimal transport plan `plan` and cost matrix `C` must be of the same size",
        )
```

src/OptimalTransport.jl
```diff
@@ -211,21 +230,26 @@ where ``H`` is the entropic regulariser, ``H(\\gamma) = -\\sum_{i, j} \\gamma_{i
...
         plan
     end
-    return dot(γ, C)
+    return vec(sum(γ .* C; dims=(1, 2)))
```
devmotion, line 252: This will lead to vectors with one element in the standard case. Maybe better
```suggestion
    return dot_matwise(γ, C)
```
and
```suggestion
dot_matwise(x::AbstractMatrix, y::AbstractMatrix) = dot(x, y)
function dot_matwise(x::AbstractArray, y::AbstractMatrix)
    xmat = reshape(x, size(x, 1) * size(x, 2), :)
    return reshape(reshape(y, 1, :) * xmat, size(x)[3:end])
end
```

src/OptimalTransport.jl
```diff
@@ -467,32 +491,51 @@ where ``\\mathrm{entropicOT}^{\\epsilon}_{C}`` denotes the entropic optimal tran
...
         throw(ArgumentError("Error: marginals are unbalanced"))
     end
-    K_all = [exp.(-C_all[i] / eps) for i in 1:length(C_all)]
+    # if batch_kernel = true, then compute all kernel reductions as matmul
+    batch_kernel = (C_all isa Matrix)
+    if batch_kernel
+        K_all = exp.(-C_all / eps)
+    else
+        K_all = [exp.(-C_all[i] / eps) for i in 1:length(C_all)]
+    end
```
devmotion, line 504: Can we again maybe compute everything in a vectorized form, i.e., use 3-dimensional arrays instead of vectors of arrays such that one can write
```suggestion
    K_all = @. exp(-C_all / eps)
```
?

test/runtests.jl
```diff
@@ -95,11 +95,33 @@ end
 
         # compute optimal transport cost (Julia implementation + POT)
         c = sinkhorn2(μ, ν, C, eps; maxiter=5_000)
-        @test c isa Float32
+        @test c isa Vector{Float32}
```
devmotion, line 98: This is what I think we should avoid (IMO it's reasonable in PythonOT since that's what POT does :shrug: but in our implementation it would be nice to avoid it).

src/OptimalTransport.jl
zsteve: Ok about this, we can certainly do so. Unlike in the case of `sinkhorn_gibbs`, even using the 3-way arrays we would still need to handle the kernel reductions separately. I'm actually beginning to think that it would be quite uncommon to solve for barycenters with multiple cost matrices (this could correspond to the input measures technically living on different spaces). So perhaps we should just require a single cost matrix (so everything lives on a common space, as per the typical usage). Then we compute in batch.

examples/basic/script.jl
```diff
@@ -206,10 +206,10 @@ heatmap(
 # the entropically regularised **barycenter** in $\mathcal{P}$ is the discrete probability
 # measure $\mu$ that solves
 # ```math
-# \inf_{\mu \in \mathcal{P}} \sum_{i = 1}^N \lambda_i \mathrm{entropicOT}^{\epsilon}_{C_i}(\mu, \mu_i)
+# \inf_{\mu \in \mathcal{P}} \sum_{i = 1}^N \lambda_i \mathrm{OT}_{\epsilon}(\mu, \mu_i)
```
devmotion, line 209: I think `\operatorname` improves spacing:
```suggestion
# \inf_{\mu \in \mathcal{P}} \sum_{i = 1}^N \lambda_i \operatorname{OT}_{\epsilon}(\mu, \mu_i)
```

examples/basic/script.jl
```diff
@@ -206,10 +206,10 @@ heatmap(
...
 # ```
-# where $\mathrm{entropicOT}^\epsilon_{C_i}(\mu, \mu_i)$ denotes the entropically regularised
-# optimal transport cost with marginals $\mu$ and $\mu_i$, cost matrix $C_i$, and entropic
+# where $\mathrm{OT}_\epsilon(\mu, \mu_i)$ denotes the entropically regularised
```
devmotion, line 211:
```suggestion
# where $\operatorname{OT}_\epsilon(\mu, \mu_i)$ denotes the entropically regularised
```

src/OptimalTransport.jl
```diff
@@ -125,26 +131,36 @@ The Gibbs kernel `K` is given by `K = exp.(- C / eps)` where `C` is the cost mat
...
+as matrices whose columns `mu[:, i]` and `nu[:, i]` correspond to pairs of histograms. 
+The output are then matrices `u` and `v` such that `u[:, i]` and `v[:, i]` are the dual variables for `mu[:, i]` and `nu[:, i]`.
 """
 function sinkhorn_gibbs(mu, nu, K; tol=1e-9, check_marginal_step=10, maxiter=1000)
-    if !(sum(mu) ≈ sum(nu))
+    if !(sum(mu; dims=1) ≈ sum(nu; dims=1))
         throw(ArgumentError("Error: mu and nu must lie in the simplex"))
     end
-
+    if size(mu, 2) != size(nu, 2)
```
devmotion, line 143: I think we might also want to allow e.g. `size(mu, 2) == 1` (this is the case even if `mu` is a vector) and `size(nu, 2) > 1` and vice versa. The main point just seems to be to ensure that `u` and `v` are sufficiently "big". I assume that one could maybe avoid adding a custom implementation here and instead use some functionality in base Julia that deals with shapes in broadcasting operations. I'll check this later today.

zsteve: Ok I've implemented one way of doing this by broadcasting. I'll try and rebase this since the code's been updated in the meantime.

zsteve:
Ok. I've made all the relevant changes and rebased to the current `master` branch. `sinkhorn` and `sinkhorn2` now support `N` to `N` and `1` to `N` batch computations, incorporating the helper functions from @devmotion . 

`sinkhorn_barycenter` now uses a common cost matrix.

zsteve:
Any chance we could merge this soon? Would be nice to start looking at batch stabilized sinkhorn etc. Also with the recent changes, I've had to keep rebasing this branch to keep up ;)

devmotion (commented):
There are some type stability issues and suboptimal implementations in the PR so we need some follow-up fixes but I guess it's a first step and I won't hold it back any longer if you need the batch support as soon as possible.

Issue #71, opened 2021-05-24
Function `emd` returning an error for an use case


davibarreira:
I'm trying to run the following case:
```julia
n = 2
distm = collect(Iterators.product(1:n,1:n));
μ = reshape(distm,n*n);
ν = reshape(distm,n*n);
C = Distances.pairwise(Distances.SqEuclidean(), μ, ν)
u = ones(n^2)/n^2
v = ones(n^2)/n^2
emd2(u,v,C,Tulip.Optimizer())
```
This is returning the following error, which I haven't been able to figure out yet:
``MathOptInterface.UnsupportedAttribute{MathOptInterface.ObjectiveFunction{MathOptInterface.ScalarAffineFunction{Int64}}}: Attribute MathOptInterface.ObjectiveFunction{MathOptInterface.ScalarAffineFunction{Int64}}() is not supported by the model.``

Any ideas?

davibarreira:
So, it looks like the error is because the matrix `C` is composed of all integers. We should either fix the documentation and say that `C` has to be float, or fix the function. What do you guys think, @devmotion and @zsteve ?

devmotion:
> I'll probably submit a PR this week adjusting the name and docs for all functions, so I'll also submit the fix.

Great, the documentation can definitely be improved (maybe take some inspiration from PythonOT?). If possible, it would be good to have separate PRs for documentation changes and bugfixes :slightly_smiling_face:

PR #72, opened 2021-05-24
Fix `emd` with integer-valued cost matrices

devmotion:
Fixes https://github.com/JuliaOptimalTransport/OptimalTransport.jl/issues/71.

@davibarreira I opened a PR right away since the fix was quite simple.

### DIFF
--- a/test/runtests.jl
+++ b/test/runtests.jl
@@ -49,11 +49,12 @@ Random.seed!(100)
         # create random cost matrix
         C = pairwise(SqEuclidean(), rand(1, M), rand(1, N); dims=2)
 
-        cost = emd2(μ, ν, C, Tulip.Optimizer())
+        # compute optimal transport map
+        P = emd(μ, ν, C, Tulip.Optimizer())
 
-        # ensure that provided map is used
-        cost2 = emd2(similar(μ), similar(ν), C, lp; plan=P)
-        @test cost2 ≈ cost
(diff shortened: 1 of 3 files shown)


davibarreira (approved):

davibarreira (approved):

davibarreira (approved):

test/runtests.jl
```diff
@@ -23,29 +23,44 @@ Random.seed!(100)
...
+    end
+
+    @testset "pre-computed plan" begin
+        # create random cost matrix
+        C = pairwise(SqEuclidean(), rand(1, M), rand(1, N); dims=2)
+
+        cost = emd2(μ, ν, C, Tulip.Optimizer())
+
+        # ensure that provided map is used
+        cost2 = emd2(similar(μ), similar(ν), C, lp; plan=P)
```
davibarreira, line 55: Since the testset changed, I guess the plan `P ` is not defined here. Hence, the test is failling.

davibarreira (changes requested):

test/runtests.jl
devmotion: Yeah, I wanted to restructure the tests a bit but apparently I messed up :smile: `lp` is not defined either...

davibarreira (approved):

davibarreira (approved):

Issue #74, opened 2021-05-25
`sinkhorn2` should return loss, not `dot(gamma, C)`


zsteve:
I noticed a problem with the current implementation of `sinkhorn2`: currently it is only returning the transport cost 
`dot(gamma, C)`, not the true Sinkhorn loss which includes the entropy penalty, i.e. `dot(gamma, C) + eps*dot(gamma, log.(gamma))`. It seems that PythonOT does the same thing (hence our unit tests work)?

What's strange is that consulting the PythonOT [docs](https://pythonot.github.io/all.html#ot.sinkhorn2), they say:
> Solve the entropic regularization optimal transport problem and return the loss.

@devmotion I think your docs in PythonOT.jl say "return the OT cost", which I interpret as being only the `dot(gamma, C)` component.

I think we should include the entropy penalty, since it's really part of what gives entropic OT its nice properties. For instance, the reason I caught this issue today was that I was trying to solve a variational problem using `dot(gamma, C)` as the loss and differentiating through `sinkhorn2`, but not having the entropy term makes the optimisation fail (I think due to lack of smoothness).

zsteve:
The solution should be straightforward: just change `dot(gamma, C)` to `dot(gamma, C) + eps*dot(gamma, log.(gamma))`. But we will need to stabilize the log for cases where `gamma` may be zero, and we take `0*log(0) = 0`.

zsteve:
> POT returns dot(gamma, C), as can be seen in the implementation (https://pythonot.github.io/_modules/ot/bregman.html#sinkhorn_knopp) and it is done on purpose it seems (PythonOT/POT#105 (comment)). So the documentation in PythonOT seems correct and I don't think the implementation should be changed since it would lead to inconsistencies with POT.

Thanks for the link. This seems strange to me, since the term `loss` should refer to the 'full' objective of an optimisation problem. Indeed, the solution offered by `sinkhorn2` is the one that minimises the loss, not cost `dot(gamma, C)`. If we want to support variational problems, then we should have some functionality that returns the loss. 

> Therefore I am not sure if there is enough reason for introducing a possibly surprising inconsistency with POT here. Maybe one could add a keyword argument that is disabled by default?

[...]

PR #75, opened 2021-05-25
Add option to return cost with regularization term

devmotion:
This PR tries to fix https://github.com/JuliaOptimalTransport/OptimalTransport.jl/issues/74 by adding a `regularization` keyword argument to `sinkhorn2`.

This change is non-breaking since, as before, by default the regularization term is not included. Moreover, this has the advantage that the behaviour of OptimalTransport.jl and POT is still consistent. Another advantage of not including the entropy term is that the cost is guaranteed to be non-negative whereas with the additional entropy term also negative values are returned (which can be problematic for downstream applications).

### DIFF
--- a/Project.toml
+++ b/Project.toml
@@ -1,18 +1,20 @@
 name = "OptimalTransport"
 uuid = "7e02d93a-ae51-4f58-b602-d97af76e3b33"
 authors = ["zsteve <stephenz@student.unimelb.edu.au>"]
-version = "0.3.1"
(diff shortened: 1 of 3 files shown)


zsteve (approved):

zsteve:
> Another advantage of not including the entropy term is that the cost is guaranteed to be non-negative whereas with the additional entropy term also negative values are returned (which can be problematic for downstream applications).

Yes, agreed. This looks good.

PR #76, opened 2021-05-26
Delete unused files

devmotion:
This PR deletes some files in the `test` directory that are not used and, in my opinion, not needed anymore: a Python script for POT and an empty image.


devmotion:
I'll merge this PR since tests pass and it does not affect the package. Moreover, the files can be recovered easily from the git history.

PR #80, opened 2021-05-27
Reorganize `sinkhorn_unbalanced` and improve convergence checks

devmotion:
This PR has a similar motivation as https://github.com/JuliaOptimalTransport/OptimalTransport.jl/pull/79 and tries to fix some spurious test errors. However, I took the opportunity and rewrote large parts of the `sinkhorn_unbalanced` implementation and added more tests.

The PR
- adds `atol` and `rtol` options for a more fine-grained convergence analysis and deprecates `tol`, similar to the other PR,
- it changes the convergence check from `0.5 * (err_a + err_b)` where `err_a` and `err_b` are the relative distance w.r.t. inf-norm of the last two iterates of `a` and `b` (same as in POT) to `isapprox(vcat(a, b), vcat(a_old, b_old); atol, rtol)` (more efficiently implemented) based on the 2-norm,
- it adds a `convergence_check` argument to adjust how often the convergence is checked,
- it renames `max_iter` to `maxiter` to be consistent with `sinkhorn`,
- it deprecates the keyword arguments `proxdiv_F1` and `proxdiv_F2` and instead uses multiple dispatch,
- it optimizes the implementation of `sinkhorn_unbalanced` and in particular gets rid of all unnecessary allocations and branches,
- it extends the tests.

Of course, one could use a different norm in the convergence check - as mentioned in the other PR, all standard norms are equivalent. However, in contrast to the marginal check in `sinkhorn` where one deals with probability measures there is no intuitive motivation for the use of the 1-norm for the scaling factors. Moreover, it seems a bit surprising to weight the errors of `a` and `b` equally even though usually they are of different dimensions.

### DIFF
--- a/src/OptimalTransport.jl
+++ b/src/OptimalTransport.jl
@@ -239,97 +239,227 @@ function sinkhorn2(μ, ν, C, ε; regularization=false, plan=nothing, kwargs...)
 end
 
 """
-    sinkhorn_unbalanced(mu, nu, C, lambda1, lambda2, eps; tol = 1e-9, max_iter = 1000, verbose = false, proxdiv_F1 = nothing, proxdiv_F2 = nothing)
+    sinkhorn_unbalanced(
+        μ, ν, C, λ1::Real, λ2::Real, ε;
+        atol=0, rtol=atol > 0 ? 0 : √eps, check_convergence=10, maxiter=1_000,
+    )
 
-Computes the optimal transport plan of histograms `mu` and `nu` with cost matrix `C` and entropic regularization parameter `eps`, 
-using the unbalanced Sinkhorn algorithm [Chizat 2016] with KL-divergence terms for soft marginal constraints, with weights `(lambda1, lambda2)`
-for the marginals `mu`, `nu` respectively.
+Compute the optimal transport plan for the unbalanced entropic regularization optimal
+transport problem with source and target marginals `μ` and `ν`, cost matrix `C` of size
+`(length(μ), length(ν))`, entropic regularization parameter `ε`, and marginal relaxation
+terms `λ1` and `λ2`.
 
-For full generality, the user can specify the soft marginal constraints ``(F_1(\\cdot | \\mu), F_2(\\cdot | \\nu))`` to the problem
-
+The optimal transport plan `γ` is of the same size as `C` and solves
 ```math
-\\min_\\gamma \\epsilon \\mathrm{KL}(\\gamma | \\exp(-C/\\epsilon)) + F_1(\\gamma_1 | \\mu) + F_2(\\gamma_2 | \\nu)
+\\inf_{\\gamma} \\langle \\gamma, C \\rangle
++ \\varepsilon \\Omega(\\gamma)
++ \\lambda_1 \\operatorname{KL}(\\gamma 1, \\mu)
++ \\lambda_2 \\operatorname{KL}(\\gamma^{\\mathsf{T}} 1, \\nu),
 ```
+where ``\\Omega(\\gamma) = \\sum_{i,j} \\gamma_{i,j} \\log \\gamma_{i,j}`` is the entropic
+regularization term and ``\\operatorname{KL}`` is the Kullback-Leibler divergence.
 
-via `math\\mathrm{proxdiv}_{F_1}(s, p)` and `math\\mathrm{proxdiv}_{F_2}(s, p)` (see Chizat et al., 2016 for details on this). If specified, the algorithm will use the user-specified F1, F2 rather than the default (a KL-divergence).
+Every `check_convergence` steps a convergence check of the error of the scaling factors
+with absolute tolerance `atol` and relative tolerance `rtol` is performed. The default
+`rtol` depends on the types of `μ`, `ν`, and `C`. After `maxiter` iterations, the
+computation is stopped.
 """
 function sinkhorn_unbalanced(
-    mu,
-    nu,
-    C,
-    lambda1,
-    lambda2,
-    eps;
-    tol=1e-9,
-    max_iter=1000,
-    verbose=false,
-    proxdiv_F1=nothing,
-    proxdiv_F2=nothing,
+    μ, ν, C, λ1::Real, λ2::Real, ε; proxdiv_F1=nothing, proxdiv_F2=nothing, kwargs...
 )
-    function proxdiv_KL(s, eps, lambda, p)
-        return @. (s^(eps / (eps + lambda)) * p^(lambda / (eps + lambda))) / s
+    if proxdiv_F1 !== nothing && proxdiv_F2 !== nothing
+        Base.depwarn(
+            "keyword arguments `proxdiv_F1` and `proxdiv_F2` are deprecated",
+            :sinkhorn_unbalanced,
+        )
+
+        # have to wrap the "proxdiv" functions since the signature changed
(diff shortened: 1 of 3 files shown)


src/OptimalTransport.jl
```diff
@@ -239,97 +239,227 @@ function sinkhorn2(μ, ν, C, ε; regularization=false, plan=nothing, kwargs...)
...
+```
+where ``\\Omega(\\gamma) = \\sum_{i,j} \\gamma_{i,j} \\log \\gamma_{i,j}`` is the entropic
+regularization term and ``\\operatorname{KL}`` is the Kullback-Leibler divergence.
 
-For full generality, the user can specify the soft marginal constraints ``(F_1(\\cdot | \\mu), F_2(\\cdot | \\nu))`` to the problem
+Every `check_convergence` steps a convergence check of the error of the scaling factors
+with absolute tolerance `atol` and relative tolerance `rtol` is performed. The default
+`rtol` depends on the types of `μ`, `ν`, and `C`. After `maxiter` iterations, the
+computation is stopped.
+"""
```
davibarreira, line 266: #79 Same comment.

src/OptimalTransport.jl
```diff
@@ -239,97 +239,227 @@ function sinkhorn2(μ, ν, C, ε; regularization=false, plan=nothing, kwargs...)
 end
 
 """
-    sinkhorn_unbalanced(mu, nu, C, lambda1, lambda2, eps; tol = 1e-9, max_iter = 1000, verbose = false, proxdiv_F1 = nothing, proxdiv_F2 = nothing)
+    sinkhorn_unbalanced(
+        μ, ν, C, λ1::Real, λ2::Real, ε;
+        atol=0, rtol=atol > 0 ? 0 : √eps, check_convergence=10, maxiter=1_000,
+    )
+
+Compute the optimal transport plan for the unbalanced entropic regularization optimal
```
davibarreira, line 247: I think it's better to call it the `unbalanced entropic regularized optimal transport problem`.

src/OptimalTransport.jl
```diff
@@ -239,97 +239,227 @@ function sinkhorn2(μ, ν, C, ε; regularization=false, plan=nothing, kwargs...)
...
+`proxdivF2!`.
 
 A pre-computed optimal transport `plan` may be provided.
 
 See also: [`sinkhorn_unbalanced`](@ref)
 """
-function sinkhorn_unbalanced2(μ, ν, C, λ1, λ2, ε; plan=nothing, kwargs...)
+function sinkhorn_unbalanced2(
+    μ, ν, C, λ1_or_proxdivF1, λ2_or_proxdivF2, ε; plan=nothing, kwargs...
+)
```
davibarreira, line 460: Shouldn't we rename to `sinkhorn_unbalanced_cost`?

src/OptimalTransport.jl
```diff
@@ -239,97 +239,227 @@ function sinkhorn2(μ, ν, C, ε; regularization=false, plan=nothing, kwargs...)
...
 
-For full generality, the user can specify the soft marginal constraints ``(F_1(\\cdot | \\mu), F_2(\\cdot | \\nu))`` to the problem
+Every `check_convergence` steps a convergence check of the error of the scaling factors
+with absolute tolerance `atol` and relative tolerance `rtol` is performed. The default
+`rtol` depends on the types of `μ`, `ν`, and `C`. After `maxiter` iterations, the
+computation is stopped.
+"""
+function sinkhorn_unbalanced(
+    μ, ν, C, λ1::Real, λ2::Real, ε; proxdiv_F1=nothing, proxdiv_F2=nothing, kwargs...
+)
```
davibarreira, line 269: Shouldn't we rename to `sinkhorn_unbalanced_plan`?

davibarreira (approved):

src/OptimalTransport.jl
devmotion: Ah yeah, this is just copied from POT, I can change it.

src/OptimalTransport.jl
devmotion: Maybe. I still think the main problem is not the name of the functions but the amount of functions - they are all doing the same thing but for slightly different problems (many even for the same) and different algorithms. So the natural approach would be to be able to dispatch both on the problem and the algorithm, which would also solve the problem that https://github.com/JuliaOptimalTransport/OptimalTransport.jl/pull/66 tries to address but doesn't fix in a general and extendable way.

devmotion: In any case, I would suggest that both renaming and reorganization of functions should be done in a separate PR.

src/OptimalTransport.jl
devmotion: Again, I'd like to avoid talking about individual iterates, how they are combined etc. As suggested in the other PR, maybe it's sufficient to point users to `isapprox`.

davibarreira: I answered in the #79.

src/OptimalTransport.jl
davibarreira: Yeah, you are right. It's better that we decide on the naming, and then submit in another PR.
" I still think the main problem is not the name of the functions but the amount of functions - they are all doing the same thing but for slightly different problems (many even for the same) and different algorithms." Why is this a problem?

devmotion: I think it's a problem since if we have different functions for every combination of problem and algorithm (such as e.g. `sinkhorn`, `sinkhorn_stabilized`, `sinkhorn_stabilized_epsscaling`, `sinkhorn_unbalanced` etc.)
- the API is unstructured and difficult to navigate for users,
- it is very difficult to compose functionality (e.g. if I would like to use epsilon scaling with the unbalanced Sinkhorn algorithm I have to write a new function instead of just composing epsilon scaling with the unbalanced algorithm),
- we completely neglect multiple dispatch which arguably is the biggest feature of Julia.

src/OptimalTransport.jl
devmotion: I noticed that also in other docstrings we use the term `entropic regularization optimal transport problem` (copied from POT as well). Do you think it should be changed there as well (in a separate PR) or should we keep `unbalanced entropic regularization optimal transport problem`? I always thought the term is a bit odd even though it's used in POT, so I would be fine with changing it in the other places as well.

davibarreira: I'd also go with changing in all the occurrences. I think it might be a typo on `POT.py` or something, because in the literature I don't remember anyone calling it like that. Usually they either call `etropic optimal transport` or `entropic regularized optimal transport`.

src/OptimalTransport.jl
devmotion: I tried to explain it more clearly and avoid redundancies. Is it OK now?

src/OptimalTransport.jl
devmotion: I went with `entropically regularized OT` now, I've seen this as well and I think I prefer it over `entropic regularized OT`.

src/OptimalTransport.jl
davibarreira: Ok

src/OptimalTransport.jl
davibarreira: Ok. I'm fine with both :)

zsteve (approved):

PR #81, opened 2021-05-27
New release

devmotion:


devmotion:
@davibarreira I can't merge this PR, it seems we really have to fix the test issues first. I restarted the tests multiple times but they fail every time (with slightly different errors).

PR #82, opened 2021-05-28
Add Zenodo badge

devmotion:
This PR adds a badge with the DOI of the package and a link to Zenodo.


devmotion:
Test error is unrelated, it's an upstream issue caused by some deprecations in the latest unstable version of Julia.

PR #88, opened 2021-05-29
Add implementation of discrete 1D optimal transport

devmotion:


src/exact.jl
```diff
@@ -0,0 +1,319 @@
...
+    νsupport = support(ν)
+    cost = sum(w * c(μsupport[i], νsupport[j]) for (i, j, w) in zip(I, J, W))
+
+    return cost
+end
+
+# fallback: compute cost matrix (probably often faster to compute cost from scratch)
+function _ot_cost(c, μ::DiscreteNonParametric, ν::DiscreteNonParametric, plan)
+    return dot(plan, StatsBase.pairwise(c, support(μ), support(ν)))
+end
```
davibarreira, line 319: Did your benchmark show that this `function _ot_cost( c, μ::DiscreteNonParametric, ν::DiscreteNonParametric, plan::SparseMatrixCSC)` was more efficient than the fallback using ` dot(plan, StatsBase.pairwise(c, support(μ), support(ν)))`? Because when I was testing my code, I kept getting better results for the latter, even for sparse matrices.

src/exact.jl
```diff
@@ -0,0 +1,319 @@
...
+    state = if μiter < max_iter
+        # move forward in the source histogram
+        (i + 1, j, iterate(d.mu, μstate), (diff, νstate))
+    else
+        # move forward in the target histogram
+        (i, j + 1, (diff, μstate), iterate(d.nu, νstate))
+    end
+
+    return iter, state
+end
```
davibarreira, line 218: Does creating an iterator improves the speed of the code?

src/exact.jl
```diff
@@ -0,0 +1,319 @@
...
+where ``F_\\mu`` is the cumulative distribution function of `μ` and ``F_\\nu^{-1}`` is the
+quantile function of `ν`.
+
+See also: [`ot_cost`](@ref), [`emd`](@ref)
+"""
+function ot_plan(c, μ::ContinuousUnivariateDistribution, ν::UnivariateDistribution)
+    # Use T instead of γ to indicate that this is a Monge map.
+    T(x) = quantile(ν, cdf(μ, x))
+    return T
+end
```
davibarreira, line 127: I just realized that we can improve this in a way that the code could actually handle distributions mu with a mix of continuous and discrete. The way it's now, we need mu to be continuous. Since the other function for 1D OT only deals with finite discrete measures, we are leaving it out measure such as Poisson (which is discrete, but with infinite support).
Although, I guess we can add it in another PR and merge this now, since an efficient way of computing 1D OT is very useful.

davibarreira (approved):

src/exact.jl
devmotion: No, but I'll perform some benchmarks. I assume it depends on the size of mu and nu - the benefit of the sparse matrix and avoiding all pairwise computations should be higher if one deals with large matrices. Additionally, probably also the sparse approach performs better with non-standard `c` since `pairwise` is optimized for `Euclidean` and `SqEuclidean`.

src/exact.jl
devmotion: The iterator per se does not improve performance, it is the same logic as in the explicit implementation. However, it removes the type instability and allows one to compute the cost without ever allocating an array for the plan in an easy way.

src/exact.jl
devmotion: Such cases require a new/different function since the plan is not a Monge map anymore. I agree that this should be done in a separate PR.

devmotion:
:eyes: I assume the test errors are caused by https://github.com/PythonOT/POT/issues/169

src/exact.jl
devmotion: I did some benchmarks: https://gist.github.com/devmotion/94ff735b13bab55ed051e8db8c906380

Short summary: For 10-dimensional mu and nu, the `pairwise` evaluation is slightly faster than the sparse iteration with `SqEuclidean` but already 10% slower with `Euclidean`. For 15-dimensional mu and nu, also with `SqEuclidean` the iteration is approx 30% faster than the `pairwise` evaluation. Of course, the difference increases with increasing dimension: For 200-dimensional mu and nu the sparse iteration is > 90% faster for both `Euclidean` and `SqEuclidean`.

davibarreira (approved):

PR #89, opened 2021-05-30
Reorganize sinkhorn barycenter

zsteve:
This PR updates `sinkhorn_barycenter` to use the same conventions for argument and variable names as `sinkhorn_gibbs` and related functions.


src/OptimalTransport.jl
```diff
@@ -715,33 +715,82 @@ Returns the entropically regularised barycenter of the `μ`, i.e. the histogram
...
+    C,
+    ε,
+    w;
+    tol=nothing,
+    atol=tol,
+    rtol=nothing,
+    check_marginal_step=nothing,
+    check_convergence=10,
+    max_iter=nothing,
+    maxiter=1_000,
```
devmotion, line 729: This should be
```suggestion
    max_iter=nothing,
    maxiter=max_iter,
```
otherwise `max_iter` is just ignored. And then one has to use
```julia
_maxiter = maxiter === nothing ? 1_000 : maxiter
```
in the iteration.

src/OptimalTransport.jl
```diff
@@ -715,33 +715,82 @@ Returns the entropically regularised barycenter of the `μ`, i.e. the histogram
...
+        Base.depwarn(
+            "keyword argument `max_iter` is deprecated, please use `maxiter`",
+            :sinkhorn_barycenter,
+        )
+    end
+    if !isapprox(extrema(sum(μ; dims=1))...)
+        throw(ArgumentError("Error: input marginals must have the same mass"))
+    end
+
+    T = float(Base.promote_eltype(μ, C))
```
devmotion, line 753: It is better to compute `K` first and then use
```suggestion
    T = float(Base.promote_eltype(μ, K))
```

src/OptimalTransport.jl
```diff
@@ -715,33 +715,82 @@ Returns the entropically regularised barycenter of the `μ`, i.e. the histogram
...
-        v = μ ./ (K' * u)
-        a = ones(size(u, 1))
-        a = prod((K * v)' .^ w; dims=1)'
-        u = a ./ (K * v)
-        if n % check_marginal_step == 0
+    # norm of inputs for convergence check
+    norm_μ = sum(abs, μ; dims=1)
+    isconverged = false
+    for iter in 1:maxiter
+        mul!(tmp, K', u)
```
devmotion, line 768: This could be skipped if a check was performed in the previous step and it would have been saved (similar to `sinkhorn_gibbs`).

src/OptimalTransport.jl
```diff
@@ -715,33 +715,82 @@ Returns the entropically regularised barycenter of the `μ`, i.e. the histogram
...
+        @. v = μ / tmp
+        mul!(tmp, K, v)
+        a .= prod(tmp' .^ w; dims=1)'
+        @. u = a / tmp
```
devmotion, line 772: This could be simplified to
```suggestion
        mul!(tmp, K, v)
        tmp2 .= tmp .^ w'
        prod!(a, tmp2)
        @. u = a / tmp
```
where `tmp2` is an array of appropriate size.

src/OptimalTransport.jl
```diff
@@ -715,33 +715,82 @@ Returns the entropically regularised barycenter of the `μ`, i.e. the histogram
...
+        if iter % check_convergence == 0
             # check marginal errors
-            err = maximum(abs.(μ .- v .* (K' * u)))
-            @debug "Sinkhorn algorithm: iteration $n" err
-            if err < tol
-                converged = true
+            mul!(tmp, K', u)
+            norm_diff = sum(abs, @. μ - v * tmp; dims=1)
```
devmotion, line 776: Similar to `sinkhorn_gibbs` (at least in the new PR) one could avoid the allocation of new arrays in every check here.

Issue #86, opened 2021-05-29
ForwardDiff errors on differentiating through the output of `sinkhorn2`


zsteve:
I'm currently trying to autodiff through `sinkhorn2` via Optim.jl, but I'm running into the following error:

```julia
julia> opt_primal = optimize(u -> f_primal(softmax(u), ε, K, interp_frac), zeros(size(μ0)), LBFGS(), Optim.Options(store_trace = true, 
show_trace = true, iterations = 250); autodiff = :forward)
ERROR: MethodError: no method matching Float64(::ForwardDiff.Dual{ForwardDiff.Tag{var"#4#5", Float64}, Float64, 11})
Closest candidates are:
  (::Type{T})(::Real, ::RoundingMode) where T<:AbstractFloat at rounding.jl:200
  (::Type{T})(::T) where T<:Number at boot.jl:760
  (::Type{T})(::AbstractChar) where T<:Union{AbstractChar, Number} at char.jl:50
  ...               
Stacktrace:    
  [1] convert(#unused#::Type{Float64}, x::ForwardDiff.Dual{ForwardDiff.Tag{var"#4#5", Float64}, Float64, 11})
    @ Base ./number.jl:7
  [2] setindex!
    @ ./array.jl:841 [inlined]
  [3] setindex!
    @ ./multidimensional.jl:639 [inlined]
  [4] macro expansion
    @ ./broadcast.jl:984 [inlined]
  [5] macro expansion
    @ ./simdloop.jl:77 [inlined]
  [6] copyto!
    @ ./broadcast.jl:983 [inlined]
  [7] copyto!
    @ ./broadcast.jl:936 [inlined]
  [8] materialize!
    @ ./broadcast.jl:894 [inlined]
  [9] materialize!
    @ ./broadcast.jl:891 [inlined]
 [10] sinkhorn_gibbs(μ::Vector{Float64}, ν::Vector{ForwardDiff.Dual{ForwardDiff.Tag{var"#4#5", Float64}, Float64, 11}}, K::Matrix{Float
64}; tol::Nothing, atol::Nothing, rtol::Nothing, check_marginal_step::Nothing, check_convergence::Nothing, maxiter::Int64)
    @ OptimalTransport ~/OptimalTransport.jl/src/OptimalTransport.jl:194
 [11] sinkhorn_gibbs
    @ ~/OptimalTransport.jl/src/OptimalTransport.jl:161 [inlined]
[...]

PR #90, opened 2021-05-30
Improve `sinkhorn_gibbs`

devmotion:
This PR improves `sinkhorn_gibbs` and fixes https://github.com/JuliaOptimalTransport/OptimalTransport.jl/issues/86.

### DIFF
--- a/src/OptimalTransport.jl
+++ b/src/OptimalTransport.jl
@@ -191,11 +191,19 @@ function sinkhorn_gibbs(
     u = similar(μ, T, size(μ, 1), size2...)
     v = similar(ν, T, size(ν, 1), size2...)
     fill!(v, one(T))
-    Kv = similar(u)
-    mul!(Kv, K, v)
 
-    norm_μ = sum(abs, μ; dims=1) # for convergence check
+    # for convergence check
+    Kv = similar(u)
     tmp = similar(u)
+    norm_μ = μ isa AbstractVector ? sum(abs, μ) : sum(abs, μ; dims=1)
+    if u isa AbstractMatrix
+        tmp2 = similar(u)
+        norm_uKv = similar(u, 1, size2...)
+        norm_diff = similar(u, 1, size2...)
+        _isconverged = similar(u, Bool, 1, size2...)
+    end
+
+    mul!(Kv, K, v)
     isconverged = false
     check_step = check_convergence === nothing ? 10 : check_convergence
     for iter in 1:maxiter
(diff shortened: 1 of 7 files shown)


devmotion:
It would be great if you could comment on the PR, I think it would be good to fix the AD issue and to improve the sinkhorn implementation.

test/entropic.jl
```diff
@@ -80,52 +103,51 @@ Random.seed!(100)
...
-            @test c_all ≈ c_pot rtol = 1e-6
-
-            γ_all = sinkhorn(μ[:, 1], ν, C, eps; maxiter=5_000)
-            γ_pot = [POT.sinkhorn(μ[:, 1], ν[:, i], C, eps; numItermax=5_000) for i in 1:d]
-            @test all([isapprox(γ_all[:, :, i], γ_pot[i]; rtol=1e-6) for i in 1:d])
-
-            γ_all = sinkhorn(μ, ν[:, 1], C, eps; maxiter=5_000)
-            γ_pot = [POT.sinkhorn(μ[:, i], ν[:, 1], C, eps; numItermax=5_000) for i in 1:d]
-            @test all([isapprox(γ_all[:, :, i], γ_pot[i]; rtol=1e-6) for i in 1:d])
+            # compute gradients with respect to source and target marginals separately and
```
zsteve, line 137: I'll see if I can make a separate PR to check the correctness of these gradients.

src/OptimalTransport.jl
```diff
@@ -87,43 +83,66 @@ function sinkhorn_gibbs(
...
+
+        # compute next iterate
+        u .= μ ./ Kv
+        mul!(v, K', u)
+        v .= ν ./ v
+        mul!(Kv, K, v)
+
+        # check source marginal
+        # always check convergence after the final iteration
+        if to_check_step <= 0 || iter == maxiter
```
zsteve, line 128: is there a reason that we stop using `iter % check_step == 0`?

devmotion: Yes, modulo operations are very expensive, the new version is a lot faster. A simple benchmark:
```julia
julia> using BenchmarkTools

julia> function f(n, m)
           x = 0
           for i in 1:n
              if i % m == 0
                  x += 1
              end
           end
           return x
       end
f (generic function with 1 method)

julia> function g(n, m)
           to_m = m
           x = 0
           for i in 1:n
               to_m -= 1
               if to_m == 0
                   x += 1
                   to_m = m
               end
           end
           return x
       end
g (generic function with 1 method)

julia> @benchmark f($(100_000), $(10))
BenchmarkTools.Trial: 
  memory estimate:  0 bytes
  allocs estimate:  0
  --------------
  minimum time:     267.902 μs (0.00% GC)
  median time:      273.221 μs (0.00% GC)
  mean time:        276.168 μs (0.00% GC)
  maximum time:     468.334 μs (0.00% GC)
  --------------
  samples:          10000
  evals/sample:     1

julia> @benchmark g($(100_000), $(10))
BenchmarkTools.Trial: 
  memory estimate:  0 bytes
  allocs estimate:  0
  --------------
  minimum time:     57.314 μs (0.00% GC)
  median time:      63.209 μs (0.00% GC)
  mean time:        63.485 μs (0.00% GC)
  maximum time:     95.818 μs (0.00% GC)
  --------------
  samples:          10000
  evals/sample:     1
```

devmotion:
What do you think, @zsteve, is it good or should anything be changed?

davibarreira:
Sorry, I saw that my review was requested, but I haven't been able to do it yet. I still have to take a better look at the code. Feel free to merge without my input, cause I'll probably take a bit longer to review.

zsteve:
> What do you think, @zsteve, is it good or should anything be changed?

This looks good!

PR #85, opened 2021-05-29
Optimal Transport for Multivariate Gaussians

davibarreira:
New PR for the Optimal Transport between Multivariate Gaussians.

This PR is very similar to the previous one (#46), but uses the implementation of Bures distance from @devmotion (https://github.com/devmotion/CalibrationErrorsDistributions.jl/blob/main/src/distances/bures.jl).

A point of contention: I was not able to figure out an efficient way of implementing the cost function when the transport plan is provided. The method I found was using cubature, which is not more efficient then just calculating using the Bures distance. I'm open to suggestions.

About the tests. I did not find any implementations of the Mv Gaussians distance in `POT.py` in order to compare the results. Hence, I did a fist case with translations only (same covariance), which has closed form solution. Then, I did a 2D case using a grid approximation and varying the covariance.

PR #91, opened 2021-06-01
Add `wasserstein` and `squared2wasserstein`

devmotion:
Add `wasserstein` and `squared2wasserstein` functions that compute the p-Wasserstein distance and the squared 2-Wasserstein distance.

https://github.com/JuliaOptimalTransport/OptimalTransport.jl/pull/88 and https://github.com/JuliaOptimalTransport/OptimalTransport.jl/pull/85 will add support for more types of marginals.

### DIFF
--- a/src/OptimalTransport.jl
+++ b/src/OptimalTransport.jl
@@ -17,9 +17,12 @@ export emd, emd2
 export sinkhorn_stabilized, sinkhorn_stabilized_epsscaling, sinkhorn_barycenter
 export sinkhorn_unbalanced, sinkhorn_unbalanced2
 export quadreg
-export ot_cost, ot_plan
+export ot_cost, ot_plan, wasserstein, squared2wasserstein
 
 const MOI = MathOptInterface
+
+include("exact.jl")
+include("wasserstein.jl")
 
 dot_matwise(x::AbstractMatrix, y::AbstractMatrix) = dot(x, y)
 function dot_matwise(x::AbstractArray, y::AbstractMatrix)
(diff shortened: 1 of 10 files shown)


src/OptimalTransport.jl
```diff
@@ -17,10 +17,13 @@ export emd, emd2
 export sinkhorn_stabilized, sinkhorn_stabilized_epsscaling, sinkhorn_barycenter
 export sinkhorn_unbalanced, sinkhorn_unbalanced2
 export quadreg
-export ot_cost, ot_plan
+export ot_cost, ot_plan, wasserstein, squared2wasserstein
```
davibarreira, line 21: Why do you think a `squared2wasserstein` is necessary? Why not only `wasserstein`? The `ot_cost` would already calculate the `squared` case if the user wished so, or he could just square the output of the `wasserstein

src/exact.jl
```diff
@@ -0,0 +1,89 @@
...
+
+The optimal transport plan solves
+```math
+\\inf_{\\gamma \\in \\Pi(\\mu, \\nu)} \\int c(x, y) \\, \\mathrm{d}\\gamma(x, y)
+```
+where ``\\Pi(\\mu, \\nu)`` denotes the couplings of ``\\mu`` and ``\\nu``.
+
+See also: [`ot_cost`](@ref)
+"""
+ot_plan
```
davibarreira, line 15: Why defining this docstring like this? Isn't it redundant since a very similar docstring is present in the following functions?

test/wasserstein.jl
```diff
@@ -0,0 +1,81 @@
...
+            @test OptimalTransport.prt(x, Val(p)) ≈ x^(1 / p)
+        end
+    end
+
+    @testset "wasserstein" begin
+        μ = Normal(randn(), randexp())
+        ν = Normal(randn(), randexp())
+        for p in (1, 2, 3, randexp()), metric in (Euclidean(), TotalVariation())
+            # without additional keyword arguments
+            w = wasserstein(μ, ν; p=Val(p), metric=metric)
```
davibarreira, line 41: is it necessary to use `p=Val(p)` when using the `wasserstein` function?

davibarreira (approved):
All good. But I wonder if `squared2wasserstein` is really necessary.

src/OptimalTransport.jl
devmotion: The squared version is useful when working with e.g. normal distributions. And it is quite expensive to first take the square root and then square the result again. Therefore also e.g. Distances contains a dedicated `SqEuclidean` distance and Distributions a `pdfL2squared`.

test/wasserstein.jl
devmotion: Yes, otherwise one can't dispatch on `p` to use optimized cost functions such as `SqEuclidean`.

src/exact.jl
devmotion: The main motivation is to explain what `ot_cost` and `ot_plan` are designed for without focusing on the special cases such as e.g. for univariate distributions or discrete 1D measures. The common practice is to only define docstrings for the special cases if they are really necessary and e.g. have special keyword arguments or return types.

davibarreira: I see. Perhaps in another PR, we should reduce some of the repeated info in the other repeated functions.

test/wasserstein.jl
davibarreira: Hmm. I understand, although it's a bit unfortunate, cause it's kind of odd.

devmotion: In the latest commit I added support for `Real`-valued `p`s and a note in the docstring that explains that sometimes `Val(p)` might be more efficient.

davibarreira (approved):

davibarreira (approved):

PR #92, opened 2021-06-02
Sinkhorn Divergence

davibarreira:
Implemented the function `sinkhorn_divergence` which is an unbiased version of `sinkhorn` and that is also a metric in the space of probability spaces. This function is similar to the `ot.bregman.empirical_sinkhorn_divergence` from `POT.py`.

The tests required the use of `PyCall`, because `ot.bregman.empirical_sinkhorn_divergence` is not supported on `PythonOT.jl`.


devmotion:
> The tests required the use of PyCall, because ot.bregman.empirical_sinkhorn_divergence is not supported on PythonOT.jl

It would be good to add it to PythonOT.

davibarreira:
> > The tests required the use of PyCall, because ot.bregman.empirical_sinkhorn_divergence is not supported on PythonOT.jl
> 
> It would be good to add it to PythonOT.

I'll see how this can be done,

davibarreira:
Submitted a PR to the `PythonOT.jl` package. Once the PR goes through there, I'll update this PR here in order to remove the `PyCall` dependency on the  test.

src/OptimalTransport.jl
```diff
@@ -785,4 +786,68 @@ function quadreg(mu, nu, C, ϵ; θ=0.1, tol=1e-5, maxiter=50, κ=0.5, δ=1e-5)
...
+"""
+    sinkhorn_divergence(c, μ::DiscreteNonParametric, ν::DiscreteNonParametric, ε; regularization=false, kwargs...)
+
+Compute the Sinkhorn Divergence between finite discrete
+measures `μ` and `ν` with respect to a cost function `c`
+and entropic regularization parameter `ε`.
+
+The Sinkhorn Divergence is computed as:
+```math
+S_{c,ε}(μ,ν) := OT_{c,ε}(μ,ν) - \\frac{1}{2}(OT_{c,ε}(μ,μ) + OT_{c,ε}(ν,ν)),
```
zsteve, line 798: Could we use `\\operatorname{OT}` here instead?

zsteve: Same for `\\operatorname{S}`

src/OptimalTransport.jl
```diff
@@ -785,4 +786,68 @@ function quadreg(mu, nu, C, ϵ; θ=0.1, tol=1e-5, maxiter=50, κ=0.5, δ=1e-5)
...
+    )
+    OTνν = sinkhorn2(
+        ν.p,
+        ν.p,
+        pairwise(c, ν.support; symmetric=true),
+        ε;
+        regularization=regularization,
+        kwargs...,
+    )
+    return max(0, OTμν - 0.5 * (OTμμ + OTνν))
```
zsteve, line 850: Is there a reason we take `max(0, ...)` here?

src/OptimalTransport.jl
```diff
@@ -785,4 +786,68 @@ function quadreg(mu, nu, C, ϵ; θ=0.1, tol=1e-5, maxiter=50, κ=0.5, δ=1e-5)
...
+[^FeydyP19]: Jean Feydy, Thibault Séjourné, François-Xavier Vialard, Shun-ichi
+Amari, Alain Trouvé, and Gabriel Peyré. Interpolating between op-
+timal transport and mmd using sinkhorn divergences. In The 22nd In-
+ternational Conference on Artificial Intelligence and Statistics, pages
+2681–2690. PMLR, 2019.
+
+See also: [`sinkhorn2`](@ref)
+"""
+function sinkhorn_divergence(
+    c,
```
zsteve, line 819: Could we stick to the calling convention of `sinkhorn` here, i.e. `(\mu, \nu, C, \varepsilon)` in that order.

davibarreira: I thought we were going to revert all functions to `(c,\mu,\nu ...)`, like the implementations of `ot_cost` and `ot_plan`. Also, I think we should use lowercase `c` for the cost function, and uppercase `C` for cost matrix.

src/OptimalTransport.jl
davibarreira: To guarantee that the value is larger or equal to 0. The same is also present in `POT.py`.

src/OptimalTransport.jl
davibarreira: Yeah, better to add the `\\operatorname`. Will do.

src/OptimalTransport.jl
zsteve: Ah I see. Makes sense, which PR is this? Sorry I've not kept up with all of them :P

src/OptimalTransport.jl
zsteve: Makes sense. It doesn't come from the 'math' though, I assume any negative values would be due to numerical issues (i.e. any negative values would be very small?)

src/OptimalTransport.jl
davibarreira: Not your fault at all! I thought we had settle on it in the issue #63 , but now I see it's not there. So probably me and @devmotion talked about in one of the Multivariate Normals or 1D implementations (which have almost hundreds of comments). Things are moving so fast in this package now, that I actually don't know where it is.
I'll comment in issue #63, so we can make a decision. If I remember correctly, it was actually a suggestion of @devmotion, based on the fact that one usually declares arguments that are functions in the beginning (I'm guessing this is a Julia standard or something).

devmotion, line 819: Could we implement this with the cost matrices as inputs instead? I think this would have multiple advantages:
- it would allow to reuse precomputed cost matrices, here and in other functions
- it would be guaranteed to work on the GPU (BTW GPU tests should be added if possible) whereas the default implementation of `pairwise` probably doesn't for most distances and custom functions (IIRC it uses incorrect containers and indexing which should be avoided on GPUs and is often disabled by users)

src/OptimalTransport.jl
```diff
@@ -785,4 +786,68 @@ function quadreg(mu, nu, C, ϵ; θ=0.1, tol=1e-5, maxiter=50, κ=0.5, δ=1e-5)
...
+    μ::DiscreteNonParametric,
+    ν::DiscreteNonParametric,
+    ε;
+    regularization=false,
```
devmotion, line 823: This can be removed since it's the default in `sinkhorn2`:
```suggestion
```

src/OptimalTransport.jl
devmotion: Yes, this is a standard and mentioned in the Julia documentation about ordering of function arguments since it allows one to use the `do` syntax
```julia
map(rand(20)) do x
    return x^2
end
```
instead of
```julia
map(x -> x^2, rand(20))
```
which is very convenient for longer and more complicated functions.

src/OptimalTransport.jl
```diff
@@ -785,4 +786,68 @@ function quadreg(mu, nu, C, ϵ; θ=0.1, tol=1e-5, maxiter=50, κ=0.5, δ=1e-5)
...
+    kwargs...,
+)
+    OTμν = sinkhorn2(
+        μ.p,
+        ν.p,
+        pairwise(c, μ.support, ν.support),
+        ε;
+        regularization=regularization,
```
devmotion, line 831:
```suggestion
```

src/OptimalTransport.jl
```diff
@@ -785,4 +786,68 @@ function quadreg(mu, nu, C, ϵ; θ=0.1, tol=1e-5, maxiter=50, κ=0.5, δ=1e-5)
...
+        kwargs...,
+    )
+    OTμμ = sinkhorn2(
+        μ.p,
+        μ.p,
+        pairwise(c, μ.support; symmetric=true),
+        ε;
+        regularization=regularization,
```
devmotion, line 839:
```suggestion
```

src/OptimalTransport.jl
```diff
@@ -785,4 +786,68 @@ function quadreg(mu, nu, C, ϵ; θ=0.1, tol=1e-5, maxiter=50, κ=0.5, δ=1e-5)
...
+        kwargs...,
+    )
+    OTνν = sinkhorn2(
+        ν.p,
+        ν.p,
+        pairwise(c, ν.support; symmetric=true),
+        ε;
+        regularization=regularization,
```
devmotion, line 847:
```suggestion
```

src/OptimalTransport.jl
devmotion, line 850: We should avoid unwanted promotions (e.g. when working with `Float32` on GPUs):
```suggestion
    return max(0, OTμν - (OTμμ + OTνν) / 2)
```

Intuitively I would have thought that the divergence can be negative if the regularization terms are included (BTW does one actually want to include them)?

src/OptimalTransport.jl
devmotion: Additionally, this would still allow to define the function implemented by just forwarding it to the one with cost matrices that are computed from the `DiscreteNonParametric` and the cost function.

devmotion: BTW it's also quite restrictive to only allow univariate messures here it seems.

test/entropic.jl
```diff
@@ -213,4 +217,26 @@ Random.seed!(100)
             @test μ_interp ≈ μ_interp_pot rtol = 1e-6
         end
     end
+
+    @testset "sinkhorn divergence" begin
+        @testset "example" begin
+            # create distributions 
+            N = 100
+            μ = DiscreteNonParametric(rand(N), ones(N) / N)
```
devmotion, line 225: Can we use random histograms?

src/OptimalTransport.jl
davibarreira: Yeah, the univariate was a mistake. I forgot that `DiscreteNonParametric` was only for the univariate case. I could implement it with a matrix cost, but it would actually require 3 matrices, so the it would be quite "ugly". It would be something like:
`sinkhorn_divergence(mu,nu,C1,C2,C3,eps)`. So I think we should stick with the cost function... I don't know much about programming for GPU, but, don't you think we could adapt this somehow?

davibarreira: I could implement another `sinkhorn_divergence` that would take the cost matrices as argument. And would could perhaps indicate that such version should be used in case the user is interested in using GPU.

devmotion: You can just define both and forward the call from the one with the cost function to the other:

> Additionally, this would still allow to define the function implemented by just forwarding it to the one with cost matrices that are computed from the DiscreteNonParametric and the cost function.

devmotion: >  such version should be used in case the user is interested in using GPU.

It could be useful without GPUs as well, e.g., if you evaluate `sinkhorn`, `sinkhorn2` or other other regularized OT distances with the same cost matrices.

In fact, it could also be useful to allow to specify pre-computed plans for the different `sinkhorn2` calls.

src/OptimalTransport.jl
davibarreira: Yeah, if the regularization terms are added, it could be negative, so you are correct that we should not allow the `regularization` keyword. For the Sinkhorn Divergence, one should probably never include the regularization terms, because the goal here is to turn the entropic ot cost into a "proper" metric in the space of probability distributions. Besides, Feydy (in the paper I cited), proved some neat properties of this metric, such as the fact that it also metrizes weak convergence.
TL;DR: I don't think one would want to ever include the regularization terms in the Sinkhorn Divergence.

For `sinkhorn2`, I saw one paper where the author suggested that the use of the cost plus regularization to behave better than when one removes it. Although, it can be negative, so it loses some of the interpretability.

davibarreira: Correct. In theory, this is a proper metric, so it would never be negative.

src/OptimalTransport.jl
davibarreira: Convinced.

test/entropic.jl
davibarreira: Yeah. I'll update.

src/OptimalTransport.jl
devmotion: Thanks for the detailed explanation! This confirms my intuition, so let's just remove the `regularization` keyword argument. Maybe then we should even check that `kwargs...` does not contain a keyword argument `regularization` and otherwise either throw an error or (maybe better) drop the `regularization` keyword argument and display a `@warn`ing. Maybe something like
```julia
....; regularization=nothing, kwargs...)
    if regularization !== nothing
        @warn "`sinkhorn_divergence` does not support the `regularization` keyword argument"
    end
```

davibarreira:
So, this current push in the PR is actually too much code for a single PR. I intend to break it down in other PR's, but, I thought it would be good to submit here the whole thing so you guys can take a look on how these new functions tie themselves to the rest of the package.
So, besides the `sinkhorn_divergence` function, I actually implemented the struct `FiniteDiscreteMeasure` and the `cost_matrix` function. Each one I'll separate in a different PR, in case we decide that they are appropriate for the package.

The `FiniteDiscreteMeasure` is a struct is a way that we can define our empirical measures and consistently write functions such as `sinkhorn2(c, mu, nu)` or `sinkhorn2(C, mu, nu)`. Right now, the package sometimes refers `mu` and `nu` only as the support, and sometimes as actual measures. So the new struct would clear things.

The `cost_matrix` is just a helper function to construct cost matrices easily. This allows us to create other versions of our functions such as `sinkhorn2` with a cost function instead of a cost matrix (similar to what I have done with sinkhorn_divergence).

@devmotion  and @zsteve , what do you think of these two new additions? Should I submit a PR for each one?

devmotion:
I think it's a good idea to move these two additions to separate PRs, then it's easier to discuss and polish them.

I also think that probably a `FiniteDiscreteMeasure` is useful, I have some additional suggestions but it's easier to take this in the corresponding PR.

I am a bit less convinced by `cost_matrix`, it seems to introduce some type instabilities but probably this could be fixed. I don't think it should be exported and become part of the official API though, but also here it's probably best to discuss this in a separate PR :slightly_smiling_face:

davibarreira:
The error in the check seems to be unrelated to the PR.

Project.toml
```diff
@@ -35,6 +35,7 @@ Random = "9a3f8284-a2c9-5f02-9a11-845980a1fd5c"
 SafeTestsets = "1bc83da4-3b8d-516f-aca4-4fe02f6d838f"
 Test = "8dfed614-e22c-5e08-85e1-65c5234f0b40"
 Tulip = "6dd1b50a-3aae-11e9-10b5-ef983d2400fa"
+PyCall = "438e738f-606a-5dbb-bf0a-cddfbfd45ab0"
```
devmotion, line 38: Not needed anymore I guess?
```suggestion
```

src/OptimalTransport.jl
```diff
@@ -4,6 +4,7 @@
 
 module OptimalTransport
 
+using LinearAlgebra: AbstractMatrix
```
devmotion, line 7:
```suggestion
```

src/entropic/sinkhorn.jl
```diff
@@ -504,3 +504,85 @@ function sinkhorn_barycenter(μ, C, ε, w; tol=1e-9, check_marginal_step=10, max
     end
     return u[:, 1] .* (K * v[:, 1])
 end
+
+"""
+    sinkhorn_divergence(
+        c,
+        μ::Union{FiniteDiscreteMeasure, DiscreteNonParametric},
+        ν::Union{FiniteDiscreteMeasure, DiscreteNonParametric},
```
devmotion, line 512: It's not really what we support, maybe just omit the type or add two docstrings (one basically referring to the other)?

src/entropic/sinkhorn.jl
```diff
@@ -504,3 +504,85 @@ function sinkhorn_barycenter(μ, C, ε, w; tol=1e-9, check_marginal_step=10, max
...
+        ε; regularization=false, plan=nothing, kwargs...
```
devmotion, line 513: It would be useful if we would restructure `sinkhorn` and introduce the different algorithms, as proposed in the other PR. Then users could also choose the stabilized algorithm or epsilon scaling here.

src/entropic/sinkhorn.jl
```diff
@@ -504,3 +504,85 @@ function sinkhorn_barycenter(μ, C, ε, w; tol=1e-9, check_marginal_step=10, max
...
+        regularization=regularization,
+        kwargs...,
+    )
+end
+
+"""
+    sinkhorn_divergence(
+        Cμν, Cμμ, Cνν,
+        μ::Union{FiniteDiscreteMeasure, DiscreteNonParametric},
+        ν::Union{FiniteDiscreteMeasure, DiscreteNonParametric},
```
devmotion, line 565: Same here.

src/entropic/sinkhorn.jl
devmotion, line 513: One should not include the regularization keyword argument since it is not supported.

```suggestion
        ε; plan=nothing, kwargs...
```

src/entropic/sinkhorn.jl
```diff
@@ -504,3 +504,85 @@ function sinkhorn_barycenter(μ, C, ε, w; tol=1e-9, check_marginal_step=10, max
...
+        ε; regularization=false, plan=nothing, kwargs...
```
devmotion, line 566:
```suggestion
        ε; plan=nothing, kwargs...
```

test/entropic/sinkhorn.jl
```diff
@@ -195,4 +196,93 @@ Random.seed!(100)
...
+            ν = OptimalTransport.discretemeasure(νsupp)
+
+            for (ε, metrics) in Iterators.product(
+                [0.1, 1.0, 10.0],
+                [
+                    (sqeuclidean, SqEuclidean()),
+                    (euclidean, Euclidean()),
+                    (totalvariation, TotalVariation()),
+                ],
+            )
```
devmotion, line 218: This is a bit uncommon, usually one would just use two `for` loops here. One can even use the short notation
```julia
for x in xs, y in ys
```

test/entropic/sinkhorn.jl
```diff
@@ -195,4 +196,93 @@ Random.seed!(100)
...
+            # create distributions 
+            n = 20
+            m = 10
+            μsupp = [rand(3) for i in 1:n]
+            νsupp = [rand(3) for i in 1:m]
+            μprobs = normalize!(rand(n), 1)
+            μ = OptimalTransport.discretemeasure(μsupp, μprobs)
+            ν = OptimalTransport.discretemeasure(νsupp)
+
+            for (ε, metrics) in Iterators.product(
```
devmotion, line 254: Same here.

test/finitediscretemeasure.jl
```diff
@@ -0,0 +1,55 @@
+using Distributions: DiscreteNonParametric
```
devmotion, line 1: This file seems unrelated to the PR?

test/runtests.jl
```diff
@@ -1,3 +1,4 @@
+using LinearAlgebra: symmetric
```
devmotion, line 1: I guess not needed?

```suggestion
```

davibarreira:
I think this one is finally ready to go.

davibarreira:
@devmotion or @zsteve, whenever you guys can, please review this PR.

PR #95, opened 2021-06-04
Finite Discrete Measure

davibarreira:
This PR has the implementation of the `FiniteDiscreteMesure` struct which I talked about in #92. It is similar to the `DiscreteNonParametric` type in `Distributions.jl`. It contains a `support`and a `p` (for probability), allowing the user to store the empirical distributions. This would allow use to write functions such as `sinkhorn2(c, mu::FiniteDiscreteMeasure, nu::FiniteDiscreteMeasure)`, without having to pass both support and probabilities separately.

p.s: I don't feel strongly about the `FiniteDiscreteMeasure` naming. Feel free to give better suggestions. I just don't think `DiscreteMeasure` is a good name, because our variable cannot have infinite support, like a Poisson distribution.
`SimpleMeasure` is another option and would be theoretically correct. `EmpiricalMeasure`and `SampleMeasure` are another option, but they are kind of ambiguous, since the measure does not need necessarily to be either empirical or from a sampling process.


src/OptimalTransport.jl
```diff
@@ -17,8 +18,11 @@ export sinkhorn, sinkhorn2
 export emd, emd2
 export sinkhorn_stabilized, sinkhorn_stabilized_epsscaling, sinkhorn_barycenter
 export sinkhorn_unbalanced, sinkhorn_unbalanced2
+export sinkhorn_divergence
```
devmotion, line 21:
```suggestion
```

src/OptimalTransport.jl
```diff
@@ -17,8 +18,11 @@ export sinkhorn, sinkhorn2
...
 export quadreg
 export ot_cost, ot_plan, wasserstein, squared2wasserstein
+export FiniteDiscreteMeasure
+export cost_matrix
```
devmotion, line 25:
```suggestion
```

test/runtests.jl
```diff
@@ -1,3 +1,4 @@
+using LinearAlgebra: symmetric
```
devmotion, line 1:
```suggestion
```

devmotion (changes requested):

src/utils.jl
```diff
@@ -54,3 +54,53 @@ function checkbalanced(x::AbstractVecOrMat, y::AbstractVecOrMat)
         throw(ArgumentError("source and target marginals are not balanced"))
     return nothing
 end
+
+struct FiniteDiscreteMeasure{X<:AbstractArray,P<:AbstractVector}
```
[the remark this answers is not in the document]
davibarreira: The issue is that in such cases, it wouldn't allow for something like `FiniteDiscreteMeasure(rand(10,2), p)`, since a matrix is not an abstract vector.

src/utils.jl
```diff
@@ -54,3 +54,53 @@ function checkbalanced(x::AbstractVecOrMat, y::AbstractVecOrMat)
...
+            DiscreteNonParametric(vec(support), p ./ P)
+        end
+    else
+        return if P ≈ 1
+            FiniteDiscreteMeasure{typeof(support),typeof(p)}(support, p)
+        else
+            FiniteDiscreteMeasure{typeof(support),typeof(p)}(support, p ./ P)
+        end
+    end
+end
```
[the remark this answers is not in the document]
davibarreira: Yeah, the sorting has to be clearly stated, I got tricked by it sometimes. About the normalization, it is a matter of convenience. It allows one to write `mu = FiniteDiscreteMeasure(rand(10,2), rand(10))`, which I find really neat. Of course, if it creates significant issues, then we should just drop it. Otherwise, I like how it allows one to write very clean code.

src/utils.jl
```diff
@@ -54,3 +54,53 @@ function checkbalanced(x::AbstractVecOrMat, y::AbstractVecOrMat)
...
+"""
+function FiniteDiscreteMeasure(support::AbstractArray)
+    p = ones(size(support)[1]) ./ size(support)[1]
+    if size(support, 2) == 1
+        return DiscreteNonParametric(vec(support), p)
+    else
+        return FiniteDiscreteMeasure{typeof(support),typeof(p)}(support, p)
+    end
+end
+
```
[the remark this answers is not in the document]
davibarreira: Again, this is not a necessary "feature", but I'd argue it's very convenient for users. I have been doing a course on OT, so pretty much every week I have created a new notebook to showcase some OT. And as a user, I really appreciate this kind of functionality in which the functions already does most of the work for me. I guess since I come from Python, I'm mostly reckless in terms of efficiency, and I'm much more user-centric.
So, if this type of stuff creates issues, I think we should drop it. Otherwise, I'd love to have it :)

src/utils.jl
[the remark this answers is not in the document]
davibarreira: I see. My bad for commenting before reading your link. I guess the `ColVecs` and `RowVecs` indeed disambiguate things.

src/utils.jl
```diff
@@ -54,3 +54,53 @@ function checkbalanced(x::AbstractVecOrMat, y::AbstractVecOrMat)
...
+struct FiniteDiscreteMeasure{X<:AbstractArray,P<:AbstractVector}
+    support::X
+    p::P
+
+    function FiniteDiscreteMeasure{X,P}(support::X, p::P) where {X,P}
+        size(support, 1) == length(p) ||
+            error("number of rows of `support` and `p` must be equal")
+        sum(p) ≈ 1 || error("`p` must sum to 1")
+        all(p .>= 0) || error("values of `p` must be greater of equal than 0")
```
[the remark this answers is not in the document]
davibarreira: I guess we could actually use the `isprobs` function from `Distributions.jl`. Just found out about it after I coded this.

davibarreira:
@devmotion , I've implemented the changes requested. But I was not sure on how to implement the `RowVecs`and `ColVecs`standard. I still have to take a better look at how `KernelFunctions.jl` does it. Is your idea for us to implement it directly in our package, or to use the functions from `KernelFunctions.jl`?

I've removed the normalization, but I still think it's a good feature to have. @zsteve , can you share your opinion on the matter?

devmotion:
> @devmotion , I've implemented the changes requested. But I was not sure on how to implement the `RowVecs`and `ColVecs`standard. I still have to take a better look at how `KernelFunctions.jl` does it. Is your idea for us to implement it directly in our package, or to use the functions from `KernelFunctions.jl`?

No, this should not be reimplemented. Users should e.g. use ArraysOfArrays.jl or KernelFunctions.jl. And hopefully soon ColVecs and RowVecs will be moved to their own package, we discussed this a while ago and IIRC there is an open issue.

> I've removed the normalization and the default assignment of equal probability to all observations in case no prob vector is provided, but I still think they are good features to have. @zsteve , what do you think?

I think the normalization is not needed for many algorithms and can easily performed manually if needed (e.g. with `LinearAlgebra.normalize!`), therefore I would not include it in this PR. Additionally, the initial implementation in the PR caused a type instability which can be avoided in this way.

I am fine with adding a default for the probabilities, you could just change `probs::AbstractVector{<:Real}` to `probs::AbstractVector{<:Real}=Fill(inv(length(support)), length(support))`.

davibarreira:
Cool, I'll do the modifications right now.

src/utils.jl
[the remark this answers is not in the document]
davibarreira: Where is this function `Fill` from ? I just know `fill`.

davibarreira:
There is some compatibility issue, but I cannot reproduce in my machine. It seems to be related to `KernelFunctions`, which I used in the test.

devmotion:
KernelFunctions requires Julia >= 1.3 (this is one reason for why we want to extract ColVecs and RowVecs to a separate package). It seems ArraysOfArrays also supports Julia 1.0, so it would be better to use it in the tests until ColVecs and RowVecs are compatible with Julia 1.0.

davibarreira:
If there are no more proposed changes, I think this is ready to be merged :D

docs/src/index.md
```diff
@@ -36,3 +36,8 @@ sinkhorn_unbalanced2
 ```@docs
 quadreg
 ```
+
+## Auxiliary functions
+```@docs
+discretemeasure
+```
```
devmotion, line 43: Maybe we could not export the function until it is actually used in some examples (e.g. in the divergence PR)? This would make it easier to change the implementation and functionality if desired - if it is exported we can't remove it and only modify it in non-breaking ways if we don't want to make a breaking release.

```suggestion
```

src/OptimalTransport.jl
```diff
@@ -20,6 +20,7 @@ export sinkhorn_stabilized, sinkhorn_stabilized_epsscaling, sinkhorn_barycenter
 export sinkhorn_unbalanced, sinkhorn_unbalanced2
 export quadreg
 export ot_cost, ot_plan, wasserstein, squared2wasserstein
+export discretemeasure
```
devmotion, line 23: Same here:

```suggestion
```

src/utils.jl
```diff
@@ -54,3 +54,55 @@ function checkbalanced(x::AbstractVecOrMat, y::AbstractVecOrMat)
...
     return nothing
 end
+
+struct FiniteDiscreteMeasure{X<:AbstractVector,P<:AbstractVector}
+    support::X
+    p::P
+
+    function FiniteDiscreteMeasure{X,P}(support::X, p::P) where {X,P}
+        length(support) == length(p) || error("length of `support` and `p` must be equal")
+        isprobvec(p) ≈ 1 || error("`p` must be a probability vector")
```
devmotion, line 64: I guess this should be

```suggestion
        isprobvec(p) || error("`p` must be a probability vector")
```
?

test/utils.jl
```diff
@@ -95,4 +96,54 @@ Random.seed!(100)
...
     end
+
+    @testset "FiniteDiscreteMeasure" begin
+        @testset "Univariate Finite Discrete Measure" begin
+            n = 100
+            m = 80
+            μsupp = rand(n)
+            νsupp = rand(m)
+            μprobs = rand(n)
+            μprobs ./= sum(μprobs)
```
devmotion, line 107:
```suggestion
            μprobs = normalize!(rand(n), 1)
```

test/utils.jl
```diff
@@ -95,4 +96,54 @@ Random.seed!(100)
...
+            ν = discretemeasure(νsupp)
+            # check if it vectors are indeed probabilities
+            @test isprobvec(μ.p)
+            @test isprobvec(probs(μ))
+            @test ν.p == ones(m) ./ m
+            @test probs(ν) == ones(m) ./ m
+
+            # check if it assigns to DiscreteNonParametric when Vector/Matrix is 1D
+            @test typeof(μ) <: DiscreteNonParametric
+            @test typeof(ν) <: DiscreteNonParametric
```
devmotion, line 119: A bit shorter:

```suggestion
            @test μ isa DiscreteNonParametric
            @test ν isa DiscreteNonParametric
```

test/utils.jl
```diff
@@ -95,4 +96,54 @@ Random.seed!(100)
...
+            @test sort(vec(νsupp)) == ν.support
+            @test sort(vec(νsupp)) == support(ν)
+        end
+        @testset "Multivariate Finite Discrete Measure" begin
+            n = 10
+            m = 3
+            μsupp = [rand(m) for i in 1:n]
+            νsupp = [rand(m) for i in 1:n]
+            μprobs = rand(n)
+            μprobs ./= sum(μprobs)
```
devmotion, line 133:
```suggestion
            μprobs = normalize!(rand(n), 1)
```

test/utils.jl
```diff
@@ -3,6 +3,9 @@ using OptimalTransport
 using LinearAlgebra
 using Random
 using Test
+using Distributions
+
+include("../src/utils.jl")
```
devmotion, line 8: I guess you added this since `discretemeasure` is not exported anymore? This should be done differently though (the location might change and/or certain parts of utils.jl might require additional packages or other types that are not available if you just include it): Remove it
```suggestion
```
and just qualify all occurrences of `discretemeasure`, i.e., use `OptimalTransport.discretemeasure` (one could also load it explicitly with `using OptimalTransport: discretemeasure` but I prefer `OptimalTransport.discretemeasure` since it clearly indicates that the function is not exported and this is how we do it for all other non-exported functions).

test/utils.jl
```diff
@@ -95,4 +98,52 @@ Random.seed!(100)
...
+    @testset "FiniteDiscreteMeasure" begin
+        @testset "Univariate Finite Discrete Measure" begin
+            n = 100
+            m = 80
+            μsupp = rand(n)
+            νsupp = rand(m)
+            μprobs = normalize!(rand(n), 1)
+
+            μ = discretemeasure(μsupp, μprobs)
+            ν = discretemeasure(νsupp)
```
devmotion, line 111:
```suggestion
            μ = OptimalTransport.discretemeasure(μsupp, μprobs)
            ν = OptimalTransport.discretemeasure(νsupp)
```

test/utils.jl
```diff
@@ -95,4 +98,52 @@ Random.seed!(100)
...
+            @test sort(vec(νsupp)) == support(ν)
+        end
+        @testset "Multivariate Finite Discrete Measure" begin
+            n = 10
+            m = 3
+            μsupp = [rand(m) for i in 1:n]
+            νsupp = [rand(m) for i in 1:n]
+            μprobs = normalize!(rand(n), 1)
+            μ = discretemeasure(μsupp, μprobs)
+            ν = discretemeasure(νsupp)
```
devmotion, line 135:
```suggestion
            μ = OptimalTransport.discretemeasure(μsupp, μprobs)
            ν = OptimalTransport.discretemeasure(νsupp)
```

test/utils.jl
davibarreira: Good to know, thanks.

devmotion (approved):
OK, let's merge this and see how we can make use of it. Since it's not exported we can iterate freely until we reach a good design :slightly_smiling_face:

PR #104, opened 2021-06-12
Cost Matrix function

davibarreira:
This PR is the separate PR for the cost matrix function, which would be a helper function in order to make writing functions as `sinkhorndivergence(c, mu, nu)` where `c` is a function.


Project.toml
```diff
@@ -35,6 +35,7 @@ Random = "9a3f8284-a2c9-5f02-9a11-845980a1fd5c"
 SafeTestsets = "1bc83da4-3b8d-516f-aca4-4fe02f6d838f"
 Test = "8dfed614-e22c-5e08-85e1-65c5234f0b40"
 Tulip = "6dd1b50a-3aae-11e9-10b5-ef983d2400fa"
+PyCall = "438e738f-606a-5dbb-bf0a-cddfbfd45ab0"
```
devmotion, line 38:
```suggestion
```

src/OptimalTransport.jl
```diff
@@ -4,6 +4,8 @@
 
 module OptimalTransport
 
+using PDMats: length
+using LinearAlgebra: AbstractMatrix
```
devmotion, line 8: Both `length` and `AbstractMatrix` are not defined in these packages but in base Julia. There should be no need to load them explicitly.
```suggestion
```

src/OptimalTransport.jl
```diff
@@ -18,8 +20,10 @@ export sinkhorn, sinkhorn2
 export emd, emd2
 export sinkhorn_stabilized, sinkhorn_stabilized_epsscaling, sinkhorn_barycenter
 export sinkhorn_unbalanced, sinkhorn_unbalanced2
+export sinkhorn_divergence
 export quadreg
 export ot_cost, ot_plan, wasserstein, squared2wasserstein
+export cost_matrix
```
devmotion, line 26: I am not sure if this should be exposed to users.

src/utils.jl
```diff
@@ -106,3 +106,66 @@ end
...
+    ν::Union{FiniteDiscreteMeasure,DiscreteNonParametric},
+)
+    if typeof(c) <: PreMetric && length(μ.support[1]) == 1
+        return pairwise(c, vcat(μ.support...), vcat(ν.support...))
+    elseif typeof(c) <: PreMetric && length(μ.support[1]) > 1
+        return pairwise(c, vcat(μ.support'...), vcat(ν.support'...); dims=1)
+    else
+        return pairwise(c, μ.support, ν.support)
+    end
+end
```
devmotion, line 148: I think we would want to restrict this to pairs of marginals with the same type of the support. Or at least the dimension should match in the case of arrays and scalars. So we would want a more fine-grained function signature.

In general, it would be better to avoid the type checks in the function definition since it makes it more difficult to extend and specialize the method. I think it would be better to just
- define a separate fallback for arbitrary `c` (the last branch)
- define a separate method for `c.:PreMetric`

Also one should avoid the splatting of support, it will lead to massive compile times and inference problems for larger arrays.

The same comments apply to the implementation below.

src/OptimalTransport.jl
davibarreira: Yeah, I remember you said this. Can you expand on why you think so? As a user, I'd like to have access to this function, since if I wanted to create the cost matrix, it would save me some time (I always misuse `pairwise` at first).

src/utils.jl
davibarreira: Hmm. So the splatter is inefficient. Now, how should one efficiently construct matrix `C` using `Distances.pairwise` ? I mean, the `StatsBase.pairwise` takes the vector of vectors and returns exactly what one wants. But the `Distances.pairwise` would require a matrix version.

src/OptimalTransport.jl
davibarreira: Also, do you think the `cost_matrix` function is useful (even if only for internal use)? Cause I remember you were not sure about it.

devmotion: The problem is that the best implementation would be to just forward everything to `pairwise` since the type of the support and the cost function should know best what to do. E.g., if you use `ColVecs` or `RowVecs` then there is no need to concatenate vectors, you can just call `pairwise` with the underlying matrix. This is also how it is implemented in KernelFunctions and it is much much more efficient than extracting and combining all columns or rows.

src/utils.jl
devmotion: As mentioned above, if you work with ColVecs or RowVecs you actually don't want to construct a matrix at all. But if you deal with an actual vector of vectors, then usually you would use e.g. `reduce(hcat, vectors_of_vectors)` to avoid splatting.

src/OptimalTransport.jl
davibarreira: Yeah, but if I construct a `FiniteDiscreteMeasure` without `KernelFunctions`, then there is no matrix attribute, this is why I wrote like that. Sorry, I don't understand your point that the best implementation would be to just forward everything to pairwise`. I mean, the reason for `cost_matrix` is that I wouldn't have to deal with the variations. For example, if my cost function is a personalized function, then `pairwise(c, mu.support, nu.support)` behaves differently than `pairwise(SqEuclidean(), mu.support, nu.support)`, which would require, for example, adding `dims=1` and transforming the support to a matrix.
What I'd like to do is to write a function that deals with all these varying cases.

src/utils.jl
davibarreira: I agree. But since we didn't require `KernelFunctions` as a dependency, I could not guarantee that `mu.support.X` would work. I'll change to your suggestion with `reduce`.

src/OptimalTransport.jl
davibarreira: Perhaps I'm not making my view point clear (or perhaps I'm the one that is not understanding your point), but I'll try to make clearer why I think this function. I want to write the `sinkhorndivergence(c, mu::FiniteDiscreteMeasure, nu::FiniteDiscreteMeasure)`. Now, if the user passes  `sinkhorndivergence(sqeuclidean, mu, nu)`, then, to construct the cost matrix I'd just write `C = pairwise(c, mu.support, nu.support)`.

If instead the user passes `sinkhorndivergence(SqEuclidean(), mu, nu)`, then I'd have to instead write `C = pairwise(c, mu.support.X, nu.support.X, dims=1)`. But since I cannot guarantee that the user user `KernelFunctions` when creating the finite measures, I have to use a method that works for both cases, hence `C = pairwise(c,reduce(hcat, mu.support), reduce(hcat, nu.support), dims=1)`.

Now, more that, the `sinkhorndivergence` also requires that I calculate the cost matrix for `mu` to itself. For `sinkhorndivergence(SqEuclidean(), mu, nu)`, the cost matrix code would be `Cmu = pairwise(sqeuclidean, mu.support, symmetry = true)`. For `SqEuclidean()` it would be `Cmu = pairwise(SqEuclidean, reduce(hcat, mu.support))`.

So what I'm trying to do with `cost_matrix` is just to avoid writing all these cases, and use `cost_matrix` to deal with that.

My bad if all of this was already clear to you.

devmotion: > If instead the user passes sinkhorndivergence(SqEuclidean(), mu, nu), then I'd have to instead write C = pairwise(c, mu.support.X, nu.support.X, dims=1). But since I cannot guarantee that the user user KernelFunctions when creating the finite measures, I have to use a method that works for both cases, hence C = pairwise(c,reduce(hcat, mu.support), reduce(hcat, nu.support), dims=1).

That's exactly my point, you don't know what type of vectors is used (e.g., whether it is a `ColVecs` or a `RowVecs`), so often `reduce(hcat, mu.support)` can be a very inefficient and suboptimal choice. If instead you would just use `pairwise(c, mu.support, nu.support)` then you could make use of the optimizations in packages such as KernelFunctions. So my main point is just that probably this is handled here on the wrong level - the packages that define e.g. `ColVecs` and `RowVecs` should define how `pairwise` is handled and make sure that it is efficient since we can't handle all possible types here.

In general though I am a bit surprised about the problems you mention with `SqEuclidean`. All the desired cases work automatically due to https://github.com/JuliaStats/Distances.jl/blob/b52f0a10017553b311a9c9eed6f96e34a5629c2f/src/generic.jl#L333-L351 (even though it is not optimized for `ColVecs` but IMO that's an issue of KernelFunctions or the separate package where they should be moved):
```julia
julia> pairwise(SqEuclidean(), rand(5), rand(5))
5x5 Matrix{Float64}:
...

julia> pairwise(SqEuclidean(), [rand(5), rand(5)], [rand(5), rand(5)])
2x2 Matrix{Float64}:
...

julia> pairwise(SqEuclidean(), ColVecs(rand(5, 2)), ColVecs(rand(5, 2)))
2x2 Matrix{Float64}:
...
```

devmotion: (`sqeuclidean` works as well but uses the fallback in StatsBase)

davibarreira: Interesting. I was getting an error, but I was using the `dims` argument. So perhaps that was the issue. If that is so, then I agree with you that the `cost_matrix` function is not necessary.

davibarreira: I'll close the PR. Thanks for the comments.

davibarreira:
It seems that this helper function was not necessary, so I'm closing this PR.

src/OptimalTransport.jl
devmotion: BTW I made a PR to Distances that would fix the SqEuclidean/sqeuclidean discrepancy: https://github.com/JuliaStats/Distances.jl/pull/224

PR #106, opened 2021-06-17
Simplify GPU test environment

devmotion:
This PR is used for debugging (and possibly fixing) the GPU errors we see lately.

I should add that the tests pass locally with CUDA 3.3.0.

### DIFF
--- a/.buildkite/pipeline.yml
+++ b/.buildkite/pipeline.yml
@@ -14,3 +14,4 @@ steps:
 
 env:
   GROUP: GPU
+  JULIA_DEBUG: CUBLAS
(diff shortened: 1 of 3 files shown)


devmotion:
OK, it seems it was caused by a bug in CUDA.jl that is already fixed in the master branch. Since a new version will be released tomorrow AFAIK, let's just wait until the new version is available.

davibarreira:
I suppose you want to merge this before merging the other PR to guarantee that they all pass the test?

devmotion:
Yes, and additionally the other PR requires a new version of NNlibCUDA (it is broken with CUDA#master).

davibarreira:
Have the CUDA issues been solved?

devmotion:
There hasn't been a new release of CUDA yet.

PR #107, opened 2021-06-18
Created CITATION.bib

davibarreira:
I've created this CITATION.bib file, so that our package works with the new `PkgCite.jl` package. We should probably automatize the process of updating this CITATION.bib file, by pulling it from Zenodo (which is what I did manually).

### DIFF
--- a/CITATION.bib
+++ b/CITATION.bib
@@ -1,0 +1,12 @@
+@software{stephen_zhang_2021_4904586,
+  author       = {Stephen Zhang and
+                  David Widmann and
+                  Davi Sales Barreira},
+  title        = {JuliaOptimalTransport/OptimalTransport.jl: v0.3.11},
+  month        = jun,
+  year         = 2021,
+  publisher    = {Zenodo},
+  version      = {v0.3.11},
+  doi          = {10.5281/zenodo.4904586},
+  url          = {https://doi.org/10.5281/zenodo.4904586}
+}


davibarreira:
By the way, I think I actually don't have the privileges to create a workflow on Actions in order to make this process automatic.

devmotion:
> By the way, I think I actually don't have the privileges to create a workflow on Actions in order to make this process automatic.

One just has to add a file in .github/workflows, there's no special permission needed.

davibarreira:
> > By the way, I think I actually don't have the privileges to create a workflow on Actions in order to make this process automatic.
> 
> One just has to add a file in .github/workflows, there's no special permission needed.

Thanks, I didn't know. Thought I had to do it via GitHub Actions interface.

devmotion:
BTW I am not sure if the Zenodo reference should/has to be added here. It seems the goal is that it is retrieved automatically from the README which would always be the most recent version and avoid this redundancy: https://github.com/SebastianM-C/PkgCite.jl/issues/1

davibarreira:
Oh, I see. So the Zenodo integration is still on progress. I think it's better to wait for that then. I'll close this PR. Thanks

devmotion:
To me this seems to be a better idea than trying to keep the reference uptodate. Updating it with Github actions e.g. if a new tag is created seems cumbersome and then the updated CITATION.bib for version 3.11 would only be part of 3.12 (since it is created after 3.11 is released). This seems strange.

davibarreira:
I'll see if I can perhaps contribute with the `PkgCite` to get the citation from zenode ;)

davibarreira:
I've submitted a PR to `PkgCite` to get badges from Zenodo. I'm just working on some modifications proposed by the author, but in a day or two everything should be up and running :D

PR #110, opened 2021-07-27
sinkhorn_barycenter refactor

zsteve:
Hi all,

I've done a quick refactor of `sinkhorn_barycenter` to now conform to conventions introduced in #100 (which I'm sorry I missed at the time due to moving), reusing code from `sinkhorn.jl` and `sinkhorn_gibbs.jl` where possible.

Problem parameters are specified in `SinkhornBarycenterSolver`, and the standard scaling algorithm corresponds to setting `alg = SinkhornBarycenterGibbs()`

I've had to define `SinkhornBarycenterGibbsCache` separately because in order to do the convergence checks, I need to keep track of the barycenter at each iteration, named `a` in the code. For the convergence check I use the existing `SinkhornConvergenceCache` struct.

Let me know if there's anything further that can be cleaned up in the code.

One thing I will note here is that quite a few parts of `solve!` can be reused between `SinkhornGibbs` and `SinkhornBarycenterGibbs` -- essentially the differences come down to the update steps and the arguments passed to the convergence check. One solution here could be to wrap the update steps into algorithm-specific calls via multiple dispatch. I haven't attempted this yet (maybe a future PR?) but I'd be keen to hear your thoughts.


devmotion:
Sorry for not replying earlier, I just came back from vacation :slightly_smiling_face: 

> One thing I will note here is that quite a few parts of `solve!` can be reused between `SinkhornGibbs` and `SinkhornBarycenterGibbs` -- essentially the differences come down to the update steps and the arguments passed to the convergence check. One solution here could be to wrap the update steps into algorithm-specific calls via multiple dispatch. I haven't attempted this yet (maybe a future PR?) but I'd be keen to hear your thoughts.

If this is possible, it could be a good way to reduce code and ensure that the implementations are consistent. However, I think this can be left for a future PR.

examples/basic/script.jl
```diff
@@ -225,7 +225,7 @@ mu = hcat(mu1, mu2)
 C = pairwise(SqEuclidean(), support'; dims=2)
 for λ1 in (0.25, 0.5, 0.75)
     λ2 = 1 - λ1
-    a = sinkhorn_barycenter(mu, C, 0.01, [λ1, λ2]; max_iter=1000)
+    a = sinkhorn_barycenter(mu, C, 0.01, [λ1, λ2], SinkhornBarycenterGibbs())
```
devmotion, line 228: Maybe it would make sense to just use `SinkhornGibbs` here? The function `sinkhorn_barycenter` already indicates that we want to compute the barycenter, it seems.

zsteve: Agreed!

PR #119, opened 2021-08-24
Refactor quadreg

zsteve:
Refactored `quadreg` to follow the conventions introduced for entropy-regularised OT. 

This PR introduces `QuadraticOT` as a base problem type, with an associated solver `QuadraticOTSolver`. The method of Lorenz et al. (2019) is re-implemented as the algorithm `QuadraticOTNewton`. 

The implementation of `solve!` currently relies on two functions `descent_dir!` and `descent_step!`, which respectively compute the descent direction (in terms of the dual variables `u, v`) and perform the descent step (currently following the Armijo step size rule in the Lorenz paper). This is actually quite general, and an alternative approach could be to rename `QuadraticOT` to a `RegularizedOT`  type for optimal transport problems with arbitrary regularizer. 

The previous implementation of `quadreg` relied on sparse matrices from `SparseArrays`, but passed between dense and sparse matrices repeatedly since transposes needed to be computed. The implementation in this PR uses dense matrices.

This could be implemented in future, but may also mean more work, since `Sinkhorn` maybe should then also inherit from `RegularizedOT`.


src/quadratic_newton.jl
```diff
@@ -0,0 +1,217 @@
+
+"""
+    Semi-smooth Newton method (Algorithm 2 of Lorenz et al. 2019) for solving quadratically regularised optimal transport
+
+    See also: [`QuadraticOTNewton`](@ref), [`quadreg`](@ref)
+
+    # References
+    
+    Lorenz, Dirk A., Paul Manns, and Christian Meyer. [Quadratically regularized optimal transport.](https://doi.org/10.1007/s00245-019-09614-w) Applied Mathematics & Optimization 83.3 (2021): 1919-1949.
```
devmotion, line 9:
```suggestion
See also: [`QuadraticOTNewton`](@ref), [`quadreg`](@ref)

# References
    
Lorenz, Dirk A., Paul Manns, and Christian Meyer. [Quadratically regularized optimal transport.](https://doi.org/10.1007/s00245-019-09614-w) Applied Mathematics & Optimization 83.3 (2021): 1919-1949.
```

src/quadratic_newton.jl
```diff
@@ -0,0 +1,217 @@
...
+    κ::K
+    δ::D
+end
+
+"""
+    QuadraticOTNewton(θ = 0.1, κ = 0.5, δ = 1e-5)
+
+    Semi-smooth Newton method (Algorithm 2 of Lorenz et al. 2019) with Armijo parameters `θ` and `κ`, and conjugate gradient regularisation `δ`. 
+    
+    See also: [`QuadraticOTNewton`](@ref), [`quadreg`](@ref)
```
devmotion, line 22:
```suggestion
Semi-smooth Newton method (Algorithm 2 of Lorenz et al. 2019) with Armijo parameters `θ` and `κ`, and conjugate gradient regularisation `δ`. 

See also: [`QuadraticOTNewton`](@ref), [`quadreg`](@ref)
```

src/quadratic_newton.jl
```diff
@@ -0,0 +1,217 @@
...
+"""
+function QuadraticOTNewton(θ=0.1, κ=0.5, δ=1e-5)
```
devmotion, line 24: Did you want to write
```suggestion
function QuadraticOTNewton(; θ=0.1, κ=0.5, δ=1e-5)
```
? Currently this leads to a stack overflow if all values are specified.

src/quadratic_newton.jl
```diff
@@ -0,0 +1,217 @@
...
+    G = cache.G
+    x = cache.x
+    M = cache.M
+    N = cache.N
+    # Armijo parameters
+    δ = solver.alg.δ
+
+    # setup intermediate variables 
+    @. γ = u + v' - C
+    @. σ = ifelse(γ ≥ 0, 1, 0)
```
devmotion, line 110: Do we need integers here? I.e., could we use
```suggestion
    @. σ = γ ≥ 0
```
?

src/quadratic_newton.jl
```diff
@@ -0,0 +1,217 @@
...
+    @. γ = NNlib.relu(γ) / eps
+
+    # setup kernel matrix G
+    G[diagind(G)[1:M]] .= vec(sum(σ; dims=2))
+    G[diagind(G)[(M + 1):end]] .= vec(sum(σ; dims=1))
+    G[1:M, (M + 1):end] .= σ
+    G[(M + 1):end, 1:M] .= σ'
+    G[diagind(G)] .+= δ # regularise cg
```
devmotion, line 118: The indexing means it does not work with GPUs (or only very slowly)?

src/quadratic_newton.jl
```diff
@@ -0,0 +1,217 @@
...
+    δv = cache.δv
+    γ = cache.γ
+
+    # Armijo parameters
+    θ = solver.alg.θ
+    κ = solver.alg.κ
+
+    # dual objective 
+    function Φ(u, v, μ, ν, C, ε)
+        return 0.5 * norm(NNlib.relu.(u .+ v' .- C))^2 - ε * dot(μ, u) - ε * dot(ν, v)
```
devmotion, line 147: Should avoid hardcoded Float64 literals:
```suggestion
        return norm(NNlib.relu.(u .+ v' .- C))^2 / 2 - ε * dot(μ, u) - ε * dot(ν, v)
```

src/quadratic_newton.jl
zsteve: My bad, yes you're right!

src/quadratic_newton.jl
zsteve: You're right, I think mainly the `diagind` is a problem here. Can use `G[1:M, 1:M] = Diagonal(...)`.

zsteve: N.B. at least for this particular algorithm the # of iterations is much smaller than e.g. with Sinkhorn. I've set the `maxiter=100` by default from playing around with numerical examples.

PR #120, opened 2021-08-25
check ForwardDiff results against analytic formula in tests

zsteve:
As mentioned in https://github.com/JuliaOptimalTransport/OptimalTransport.jl/pull/90#discussion_r644203936, this PR checks the gradients of `sinkhorn2` in its source and target measures (as obtained by ForwardDiff) against the analytic expression for the gradients.

Reference: Proposition 2.3 of Cuturi, Marco, and Gabriel Peyré. "A smoothed dual approach for variational Wasserstein problems." SIAM Journal on Imaging Sciences 9.1 (2016): 320-343.

Note that for the result of Prop 2.3 to yield a valid gradient, one needs to run Sinkhorn up til convergence (potentially many iterations). The benefit of using Sinkhorn+AD is that AD can be used to yield a valid gradient (of the suboptimal Sinkhorn objective at a fixed iteration) for relatively small number of iterations. See e.g. https://arxiv.org/pdf/1708.01955.pdf.

### DIFF
--- a/test/entropic/sinkhorn_gibbs.jl
+++ b/test/entropic/sinkhorn_gibbs.jl
@@ -157,24 +157,56 @@ Random.seed!(100)
     # https://github.com/JuliaOptimalTransport/OptimalTransport.jl/issues/86
     @testset "AD" begin
         # compute gradients with respect to source and target marginals separately and
-        # together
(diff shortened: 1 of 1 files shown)


zsteve:
Merging since this only affects tests.

PR #121, opened 2021-08-27
Quad gpu fix

zsteve:
This PR fixes type issues that prevent `quadreg` from running with CuArrays, and adds a test for that setting.
Also fixes a bug with the Armijo step search which I found for certain inputs caused the solver to get stuck.


devmotion (approved):
Looks good to me!

PR #122, opened 2021-08-27
Example: variational problems using autodiff

zsteve:
Added a new example notebook `examples/variational` demonstrating how to use the package to differentiate through the output of `sinkhorn2` and approximate solutions to PDEs as Wasserstein gradient flows.


davibarreira:
Hey @zsteve , what exactly is this optimization that you used to minimize the $\rho$ (the variational problem)? This is the only thing that it wasn't quite clear to me when reading the example.

```
function step(ρ0, τ, ε, C, G)
    opt = optimize(
        u -> G(softmax(u), ρ0, τ, ε, C),
        ones(size(ρ0)),
        LBFGS(),
        Optim.Options(; iterations=50, g_tol=1e-6);
        autodiff=:forward,
    )
    return softmax(Optim.minimizer(opt))
end
```

PR #123, opened 2021-08-28
Unify `solve!` for `SinkhornSolver` and `SinkhornBarycenterSolver`

zsteve:
I've combined `solve!` to have a unified implementation `solve!(solver::Union{SinkhornSolver, SinkhornBarycenterSolver})`. Taking a type union is not so elegant but it works since we don't have much of a type hierarchy. 

The actual steps of Sinkhorn type algorithms are now isolated into calls to:
- `init_step!(solver)` - initialise cache
- `prestep!(solver, iter)` - carried over from previously. although this _could_ be merged with `step!` below
- `step!(solver,  iter)` - one step of the scaling algorithm.

Since `solve!` now depends on both solvers, it's now defined in a separate `sinkhorn_solve.jl` file and included at the end.


examples/benchmark/script.jl
```diff
@@ -0,0 +1,43 @@
...
+
+N = 100
+μ = fill(1 / N, N)
+ν = fill(1 / N, N)
+C = rand(N, N)
+
+# Exact 
+@benchmark begin
+    emd2(μ, ν, C, Tulip.Optimizer())
+end
```
devmotion, line 18: For accurate benchmarks one should interpolate global variables. There is some explanation and example in the README: https://github.com/JuliaCI/BenchmarkTools.jl#quick-start

zsteve: Apologies, this script shouldn't even be there (I thought it was in a different local branch!). Removing in next commit, but thanks for the link!

zsteve:
Let me know if any additional changes are required before merge!

test/entropic/sinkhorn_barycenter.jl
```diff
@@ -1,6 +1,7 @@
 using OptimalTransport
 
 using Distances
+using ForwardDiff
```
devmotion, line 4: Is it actually needed?
```suggestion
```

devmotion (commented):
It's always good to reduce code duplication :+1: I'm not sure if this is the optimal design but as long as these functions are only used internally it's easy to revise and update at some point if desired.

test/entropic/sinkhorn_barycenter.jl
zsteve: Nope - sorry this seems to be related to spillover from some other branch :/ I'll take a look through files to double check.

PR #125, opened 2021-08-28
Improve setup of examples

devmotion:
This PR separates the configuration of examples and how they are built more cleanly from the rest of the documentation. Currently, at build time environments are nested and hence not cleanly separated. Moreover, with the PR examples are run in parallel.

The same setup is used e.g. in AbstractGPs.jl and KernelFunctions.jl.


zsteve (approved):
Sounds good!

PR #126, opened 2021-08-30
Fix CompatHelper

devmotion:


devmotion:
I ran CompatHelper manually with this PR to ensure that it works now: https://github.com/JuliaOptimalTransport/OptimalTransport.jl/runs/3461048105?check_suite_focus=true

PR #128, opened 2021-08-30
Update examples and reduce dependencies

devmotion:
Removes ReverseDiff, Suppressor, and LaTeXStrings and instead uses the logging standard library and raw strings. Also updates the packages.

Replaces https://github.com/JuliaOptimalTransport/OptimalTransport.jl/pull/127.


zsteve (approved):
Looks good to me!

PR #129, opened 2021-09-04
Add submodule `OptimalTransport.Dual` and examples for smoothed dual variational problems

zsteve:
This PR adds a submodule `Dual` that provides the duals of the smoothed Wasserstein loss and their gradients (so far only entropy-regularised, but could expand to L2 or group-sparse regularisations, see https://arxiv.org/abs/1710.06276). 

This kind of stuff is more geared towards variational problem use cases where OT is used as part of a loss function, rather than to 'statically' measure distances between prescribed distributions.

As an illustration of how dual approaches can be practically useful, the JKO example is updated with code that solves the dual problem (and this is much faster than using the primal approach of autodifferentiating `sinkhorn`). 

An additional example implementing Wasserstein-NMF as described in Rolet et al. (https://proceedings.mlr.press/v51/rolet16.pdf) is provided, and this demonstrates uses of the `Dual` submodule. 

P.S. apologies for the commits trying to get CIs to pass -- just figured out to set JULIA_LOAD_PATH="@" in order to prevent usage of packages from outside the active Julia environment!


docs/Project.toml
```diff
@@ -1,5 +1,6 @@
 [deps]
 Documenter = "e30172f5-a6a5-5a46-863b-614d45cd2de4"
+Literate = "98b081ad-f1c9-55d3-8b20-4c87d4299306"
```
devmotion, line 3: In the current setup on the master branch it is assumed that Literate is part of the dependencies of the example itself but not of the documentation.

examples/nmf/script.jl
```diff
@@ -0,0 +1,347 @@
...
+# For each of these problems, the dual problem can be derived (see Section 3.3 of Rolet et al. for details on how this is done). These turn out to be 
+# ```math 
+# \begin{aligned}
+# &\min_{G \in \mathbb{R}^{m \times n}} \sum_{i = 1}^n \operatorname{OT}_{\varepsilon}^*(X_i, G_i) + \rho_2 \sum_{i = 1}^{n} E^*(-(G \Lambda^\top)_i / \rho_2) \\ 
+# &\min_{G \in \mathbb{R}^{m \times n}} \sum_{i = 1}^n \operatorname{OT}_{\varepsilon}^*(X_i, G_i) + \rho_1 \sum_{i = 1}^{n} E^*(-(D^\top G)_i / \rho_1). 
+# \end{aligned}
+# ```
+# The semi-dual of entropy-regularised optimal transport loss, $\operatorname{OT}_{\varepsilon}$, is implemented as `Dual.ot_entropic_semidual` and its gradient can be computed by `Dual.ot_entropic_semidual_grad`. $E^*$ turns out to be `logsumexp`, for which we implement a wrapper function:
+#
+function E_star(x; dims=1)
```
devmotion, line 84: Maybe it's a bit more explicit without the wrapper? Or alternatively maybe use unicode characters?

examples/nmf/script.jl
```diff
@@ -0,0 +1,347 @@
...
+    return getprimal_weights(D, Optim.minimizer(opt), ρ1)
+end
+function solve_dict(X, K, ε, Λ, ρ2; alg, options)
+    opt = optimize(
+        g -> dual_obj_dict(X, K, ε, Λ, g, ρ2),
+        (∇, g) -> dual_obj_dict_grad!(∇, X, K, ε, Λ, g, ρ2),
+        zero.(X),
+        alg,
+        options;
+        autodiff=:forward,
```
devmotion, line 144: This is the default in Optim, isn't it?

devmotion: BTW, also in the other example, intuitively I would assume that forward mode AD is slower than reverse mode AD.

examples/nmf/script.jl
```diff
@@ -0,0 +1,347 @@
...
+# 
+# We visualise the  observations. 
+#
+plot(coord, X; alpha=0.1, color=:blue, title="Input data", legend=nothing)
+# 
+# We can apply NMF with a squared Frobenius loss using the NMF.jl package. We seek `k = 3` components.  This performs poorly, since the pointwise nature of the loss function cannot handle the translational noise in the data.
+# 
+k = 3
+result = nnmf(X, k; alg=:multmse)
+plot(coord, result.W; title="NMF with Frobenius loss", palette=:Set1_3)
```
devmotion, line 175: FYI you can set default settings for Plots with `default(palette=:Set1_3)` at the top only once IIRC.

examples/nmf/script.jl
```diff
@@ -0,0 +1,347 @@
...
+# ## Example: image data (MNIST)
+# Here we will download MNIST dataset using MLDatasets.jl and downscale each image to 14x14 to allow for faster runtime (since we are running on CPU). 
+#
+sizex, sizey = 28, 28
+factor = 2 # downscale factor 
+Σ = hcat([sum(I(sizex)[:, i:(i + factor - 1)]; dims=2) for i in 1:factor:sizex]...)
+sizex, sizey = sizex ÷ factor, sizey ÷ factor
+N = 256
+# must set this for Documenter
+ENV["DATADEPS_ALWAYS_ACCEPT"] = true
```
devmotion, line 236: We could just set it in the Github action.

src/utils.jl
```diff
@@ -24,6 +24,17 @@ function dot_matwise(x::AbstractArray, y::AbstractMatrix)
 end
 dot_matwise(x::AbstractMatrix, y::AbstractArray) = dot_matwise(y, x)
 
+function dot_vecwise(x::AbstractMatrix, y::AbstractMatrix)
+    return [dot(u, v) for (u, v) in zip(eachcol(x), eachcol(y))]
+end
+
+function dot_vecwise(x::AbstractMatrix, y::AbstractVector)
+    return [dot(u, y) for u in eachcol(x)]
+end
```
devmotion, line 33: I assume it is faster to compute
```suggestion
dot_vecwise(x::AbstractMatrix, y::AbstractVector) = x' * y
```

devmotion (commented):
I had a quick high-level look at the PR and added some initial comments :slightly_smiling_face: It looks already quite polished. I'll inspect it more closely within the next days. One thing that I noticed is that the function names seem quite long, maybe it is possible to shorten them without making them too cryptic?

docs/Project.toml
zsteve: Makes sense! Fixed.

examples/nmf/script.jl
zsteve: The motivation for using the wrapper function instead of just logsumexp is for better readability. With unicode, unfortunately `*` and `⋆` don't seem to be allowed in variable names...

examples/nmf/script.jl
zsteve: Yes, reverse diff is faster for this. Looks like Optim.jl doesn't allow `autodiff=:reverse` but I'll add this manually.

zsteve: > This is the default in Optim, isn't it?
No, I think the default is to use finite differences. But not necessary when the gradient is already provided. Fixed

examples/nmf/script.jl
zsteve: Thanks! Suggestion incorporated

examples/nmf/script.jl
zsteve: Done!

zsteve:
Let me know if anything further is needed for this PR!

zsteve:
Merging since there don't seem to be additional comments.

PR #130, opened 2021-09-07
fix sinkhorn2 bug for ReverseDiff

zsteve:
Using `ReverseDiff` with the current implementation of `sinkhorn2` breaks:
```julia
using OptimalTransport
using ReverseDiff
using ForwardDiff
using LogExpFunctions
import NNlib
using LinearAlgebra

N = 100 
C = rand(N, N)
ε = 0.05

μ = NNlib.softmax(zeros(N,); dims = 1)
ν = NNlib.softmax(zeros(N,); dims = 1)

ReverseDiff.gradient(x -> sinkhorn2(μ, softmax(x), C, ε; regularization = true), zero(ν))
```
yields an error
```julia
ERROR: MethodError: no method matching index_bound(::ReverseDiff.TrackedReal{Float64, Float64, Nothing}, ::ReverseDiff.TrackedReal{Float64, Float64, Nothing})
Closest candidates are:
  index_bound(::Any, ::AbstractArray{T, N}) where {T, N} at /home/syz/.julia/packages/ReverseDiff/E4Tzn/src/derivatives/propagation.jl:25
Stacktrace:
  [1] broadcast_plus(x::ReverseDiff.TrackedReal{Float64, Float64, Nothing}, y::Array{ReverseDiff.TrackedReal{Float64, Float64, Nothing}, 0}, #unused#::Type{Float64})
    @ ReverseDiff ~/.julia/packages/ReverseDiff/E4Tzn/src/derivatives/elementwise.jl:360
  [2] broadcast
    @ ~/.julia/packages/ReverseDiff/E4Tzn/src/derivatives/elementwise.jl:342 [inlined]
  [3] _materialize
    @ ~/.julia/packages/ReverseDiff/E4Tzn/src/derivatives/broadcast.jl:265 [inlined]
  [4] materialize
    @ ~/.julia/packages/ReverseDiff/E4Tzn/src/derivatives/broadcast.jl:282 [inlined]
  [5] sinkhorn2(μ::Vector{Float64}, ν::Vector{ReverseDiff.TrackedReal{Float64, Float64, ReverseDiff.TrackedArray{Float64, Float64, 1, Vector{Float64}, Vector{Float64}}}}, C::Matrix{Float64}, ε::Float64, alg::SinkhornGibbs; regularization::Bool, plan::Nothing, kwargs::Base.Iterators.Pairs{Symbol, Union{Nothing, Int64}, Tuple{Symbol, Symbol}, NamedTuple{(:atol, :check_convergence), Tuple{Nothing, Int64}}})
    @ OptimalTransport ~/OptimalTransport.jl/src/entropic/sinkhorn.jl:204
  [6] #sinkhorn2#37
    @ ~/OptimalTransport.jl/src/entropic/sinkhorn_gibbs.jl:110 [inlined]
  [7] (::var"#3#4")(x::ReverseDiff.TrackedArray{Float64, Float64, 1, Vector{Float64}, Vector{Float64}})
    @ Main ./REPL[14]:1
  [8] ReverseDiff.GradientTape(f::var"#3#4", input::Vector{Float64}, cfg::ReverseDiff.GradientConfig{ReverseDiff.TrackedArray{Float64, Float64, 1, Vector{Float64}, Vector{Float64}}})
    @ ReverseDiff ~/.julia/packages/ReverseDiff/E4Tzn/src/api/tape.jl:199
  [9] gradient(f::Function, input::Vector{Float64}, cfg::ReverseDiff.GradientConfig{ReverseDiff.TrackedArray{Float64, Float64, 1, Vector{Float64}, Vector{Float64}}}) (repeats 2 times)
    @ ReverseDiff ~/.julia/packages/ReverseDiff/E4Tzn/src/api/gradients.jl:22
 [10] top-level scope
    @ REPL[14]:1
```
The error is due to the second line of the below, which handles both cases where the output is scalar or array. This seems to cause issues with ReverseDiff.
```julia
        dot_matwise(γ, C) .+                                                                                                                                                                                                                                                   
        ε * reshape(sum(LogExpFunctions.xlogx, γ; dims=(1, 2)), size(γ)[3:end])
```
This PR instead uses a multiple dispatch on the type of `γ`, and so mitigates the error. I've updated the tests to check gradient computations using both ForwardDiff and ReverseDiff.


Project.toml
```diff
@@ -13,6 +13,7 @@ MathOptInterface = "b8f27783-ece8-5eb3-8dc8-9495eed66fee"
 NNlib = "872c559c-99b0-510c-b3b7-b6c96a88d5cd"
 PDMats = "90014a1f-27ba-587c-ab20-58faa44d9150"
 QuadGK = "1fd47b50-473d-5c70-9696-f719f8f3bcdc"
+ReverseDiff = "37e2e3b7-166d-5795-8a7a-e32c996b4267"
```
devmotion, line 16: ReverseDiff should only be a test dependency.

src/entropic/sinkhorn.jl
```diff
@@ -278,6 +278,13 @@ supported here are the same as those in the [`sinkhorn`](@ref) function.
 See also: [`sinkhorn`](@ref)
 """
 function sinkhorn2(μ, ν, C, ε, alg::Sinkhorn; regularization=false, plan=nothing, kwargs...)
+    function reg(γ::AbstractMatrix, C::AbstractMatrix, ε)
+        return ε * only(sum(LogExpFunctions.xlogx, γ; dims=(1, 2)))
```
devmotion, line 194: Only is not available on Julia < 1.4.

src/entropic/sinkhorn.jl
```diff
@@ -190,6 +190,13 @@ supported here are the same as those in the [`sinkhorn`](@ref) function.
 See also: [`sinkhorn`](@ref)
 """
 function sinkhorn2(μ, ν, C, ε, alg::Sinkhorn; regularization=false, plan=nothing, kwargs...)
+    function reg(γ::AbstractMatrix, C::AbstractMatrix, ε)
```
devmotion, line 193: Maybe this should be a separate function outside of `sinkhorn2`.

devmotion (commented):
There are some issues with the current implementation.

Another problem of the proposed fix is that it won't allow us anymore to fuse the broadcast operations: we should have use `dot_matwise(...) .+ eps .* reshape(...)` on the master branch, it seems currently an intermediate array is created if `reshape(...)` returns an array. Maybe this fixes already the ReverseDiff issue?

zsteve:
> Another problem of the proposed fix is that it won't allow us anymore to fuse the broadcast operations: we should have use dot_matwise(...) .+ eps .* reshape(...) on the master branch, it seems currently an intermediate array is created if reshape(...) returns an array. Maybe this fixes already the ReverseDiff issue?

Thanks for the insight - just checked this and indeed you're right :D It fixes the issue much more painlessly. Will revert to that and keep the tests.

devmotion (approved):
Looks good to me! I would suggest bumping the version number in the PR as well so we can make a bugfix release when it is merged.

Feel free to merge if tests pass.

PR #145, opened 2021-09-13
add davibarreira's sinkhorn_divergence with some modifications

zsteve:
This PR adds `sinkhorn_divergence` due to @davibarreira in PR #92, except I have removed the use of `FiniteDiscreteMeasure` and `DiscreteNonParametric`, etc. For now the implementation takes measures as vectors of weights, and precomputed cost matrices as supplied. A future PR could implement wrappers that take generic distributions as inputs. 

Notably, in the case where the support is fixed and there is a common cost matrix between source and target measures (as is the case on a fixed discrete domain), `sinkhorn_divergence` now takes advantage of the batched Sinkhorn computation.


devmotion (changes requested):
Nice, I wanted to update the PR this week but since I am a bit busy I'm happy that you took it up.

I have two major suggestions/requests though:
1. I think we should allow both with and without regularization. POT includes the regularization as well as Feydy et al whereas Genevay et al didn't it seems.
2. Probably we don't want to use standard Sinkhorn algorithm for the symmetric cases but instead the fixed point iterations in Feydy et al which apparently converge much faster.

davibarreira:
> This PR adds `sinkhorn_divergence` due to @davibarreira in PR #92, except I have removed the use of `FiniteDiscreteMeasure` and `DiscreteNonParametric`, etc. For now the implementation takes measures as vectors of weights, and precomputed cost matrices as supplied. A future PR could implement wrappers that take generic distributions as inputs.
> 
> Notably, in the case where the support is fixed and there is a common cost matrix between source and target measures (as is the case on a fixed discrete domain), `sinkhorn_divergence` now takes advantage of the batched Sinkhorn computation.

Thanks, @zsteve, for updating the PR. Why did you remove the use of `FiniteDiscreteMeasure`? I have to catch up to the recent changes you guys made.

davibarreira:
> Nice, I wanted to update the PR this week but since I am a bit busy I'm happy that you took it up.
> 
> I have two major suggestions/requests though:
> 
>     1. I think we should allow both with and without regularization. POT includes the regularization as well as Feydy et al whereas Genevay et al didn't it seems.
> 
>     2. Probably we don't want to use standard Sinkhorn algorithm for the symmetric cases but instead the fixed point iterations in Feydy et al which apparently converge much faster.

@devmotion what is this fixed points iterations? If I remember correctly, the paper of Feydy used Sinkhorn with fixed number of iterations. Is this what you are referring to?

devmotion:
No, I am not referring to fixed number of iterations. They use a fixed point iteration for the symmetric that is different from the standard Sinkhorn algorithm and exploits the additional structure of the problem.

zsteve:
> > This PR adds `sinkhorn_divergence` due to @davibarreira in PR #92, except I have removed the use of `FiniteDiscreteMeasure` and `DiscreteNonParametric`, etc. For now the implementation takes measures as vectors of weights, and precomputed cost matrices as supplied. A future PR could implement wrappers that take generic distributions as inputs.
> > Notably, in the case where the support is fixed and there is a common cost matrix between source and target measures (as is the case on a fixed discrete domain), `sinkhorn_divergence` now takes advantage of the batched Sinkhorn computation.
> 
> Thanks, @zsteve, for updating the PR. Why did you remove the use of `FiniteDiscreteMeasure`? I have to catch up to the recent changes you guys made.

I think that was either done earlier (not exactly sure when? or possibly never merged?). For the time being perhaps best to keep as-is (specify vectors of weights, and for empirical measures the locations are encoded in the cost matrix). More than happy to reintroduce `FiniteDiscreteMeasure` though, and can then e.g. write wrapper API.

davibarreira:
My bad, I think the implementation was `DiscreteNonParametric`.

zsteve:
> My bad, I think the implementation was `DiscreteNonParametric`.
Ah ok. then that's from Distributions.jl right?

davibarreira:
> > My bad, I think the implementation was `DiscreteNonParametric`.
> > Ah ok. then that's from Distributions.jl right?

I'm reading the code again, it has been quite some time. But for now, the way you coded seems fine. I can then submit another PR later if that's the case.

src/entropic/sinkhorn.jl
```diff
@@ -209,3 +209,102 @@ function sinkhorn2(μ, ν, C, ε, alg::Sinkhorn; regularization=false, plan=noth
...
+2681–2690. PMLR, 2019.
+See also: [`sinkhorn2`](@ref)
+"""
+function sinkhorn_divergence(
+    μ::AbstractVecOrMat,
+    ν::AbstractVecOrMat,
+    C,
+    ε,
+    alg::Sinkhorn=SinkhornGibbs();
+    regularization=nothing,
```
davibarreira, line 247: It looks like you remove the `regularization` variable, and just set it all to false. But then we cannot choose to use the value with the regularization, as in the pure `sinkhorn`.

src/entropic/sinkhorn.jl
```diff
@@ -209,3 +209,102 @@ function sinkhorn2(μ, ν, C, ε, alg::Sinkhorn; regularization=false, plan=noth
...
+"""
+function sinkhorn_divergence(
+    μ,
+    ν,
+    Cμν,
+    Cμ,
+    Cν,
+    ε,
+    alg::Sinkhorn=SinkhornGibbs();
+    regularization=nothing,
```
davibarreira, line 298: Same as above.

davibarreira (changes requested):
I think the only issue I found was with the `regularization`, which seems that it's not been used.

zsteve:
Ok, I think I've worked out the update for the symmetric Sinkhorn (Eq. 25 of Feydy et al.), but in the variables `u = exp(f/eps)`.
```
log(u) <- (log(u) + log(K*(mu*u)))/2
```
where `K` is the Gibbs kernel. This should probably (?) be implemented e.g. as a `SymmetricSinkhorn` problem through the `solve!` API. What do you think @devmotion @davibarreira

zsteve:
P.S. as part of testing autodiff with `sinkhorn_divergence`, added a short example that demonstrates Figure 1 of Feydy et al.
image

davibarreira:
> P.S. as part of testing autodiff with `sinkhorn_divergence`, added a short example that demonstrates Figure 1 of Feydy et al.
> image

Just to guarantee that we are in the same page, we are talking about the paper "Interpolating between optimal transport and mmd using sinkhorn divergences", correct?

zsteve:
yes, this one: http://proceedings.mlr.press/v89/feydy19a/feydy19a.pdf

devmotion:
> Ok, I think I've worked out the update for the symmetric Sinkhorn (Eq. 25 of Feydy et al.), but in the variables `u = exp(f/eps)`.
> 
> ```
> log(u) <- (log(u) + log(K*(mu*u)))/2
> ```
> 
> where `K` is the Gibbs kernel. This should probably (?) be implemented e.g. as a `SymmetricSinkhorn` problem through the `solve!` API. What do you think @devmotion @davibarreira

I think `SymmetricSinkhorn` might be a good approach but I'd suggest not exporting it (I don't think it's very useful as a standalone algorithm). I would assume that it would be numerically more stable to perform the calculations in log space, i.e., to work with `f` (or `f / eps`) instead of `u` (if this is what you suggested). I.e. to use
```math
f <- (f .- eps .* vec(logsumexp(log.(mu) .+ f ./ eps .- C' ./eps; dims=1))) ./ 2
```
or
```math
g <- (g .- vec(logsumexp(log.(mu) .+ g .- C' ./eps; dims=1))) ./ 2
```
where `g = f / eps`.

zsteve:
Agree with not exporting. In terms of the fixed point iteration, the current `sinkhorn_gibbs` algorithm doesn't do computation in log-space. I guess the upshot of that approach is that you can use BLAS for the multiplication with the kernel matrix `K`. Probably doesn't hurt to implement both, though.

zsteve:
I've implemented `SymmetricSinkhorn` now and changed `sinkhorn_divergence` to take advantage of it for the correction terms. Running a few tests with debug output shown reveals that indeed SymmetricSinkhornGibbs converges an order of magnitude fewer iterations than SinkhornGibbs, so thanks @devmotion  for pointing that out!

The scheme derived in Feydy et al. uses a reference measure `\mu \otimes \nu`, whereas the reference measure used by `sinkhorn2`, etc. is `\ones \otimes \ones`. From Proposition 4.2 of Computational Optimal Transport (G. Peyre), the choice of reference measure is irrelevant but just amounts to a choice of scaling. In the Sinkhorn divergence formula, switching from one reference measure to another produces terms that cancel out.

Where `u = v` are the dual variables used by the standard`SinkhornGibbs`, the symmetric-case fixed point iteration scheme of Feydy should be
```
u <- sqrt.( u .* mu/(K * u))
```
where `u = exp.(f/eps)` and `f` is as in Feydy et al. 

Another thing that I notice is that `sinkhorn2` with `regularization = true` computes in fact the full transport plan `\gamma`. When we have the dual variables `(u, v)`, one can in fact compute the loss (= cost + regulariser) by only doing matrix-vector products.
```
sinkhorn_loss = eps*(dot(u .* log.(u), K*v) + dot(v.*log.(v), K'*u))
```
When `regularizer = true`, this should be a more efficient computation as it avoids allocation of the transport plan matrix. As it is currently implemented, we still incur a vector allocation from `K'*u`.

zsteve:
> Nice, I wanted to update the PR this week but since I am a bit busy I'm happy that you took it up.
> 
> I have two major suggestions/requests though:
> 
> 1. I think we should allow both with and without regularization. POT includes the regularization as well as Feydy et al whereas Genevay et al didn't it seems.
This is now addressed by the `regularization` argument, which is no longer ignored. The unit tests check for both cases, `regularization in (true, false)`
> 2. Probably we don't want to use standard Sinkhorn algorithm for the symmetric cases but instead the fixed point iterations in Feydy et al which apparently converge much faster.
This is now addressed by the `SymmetricSinkhorn` algorithm.

src/entropic/sinkhorn.jl
zsteve: Sorry, this is fixed now!

src/entropic/sinkhorn.jl
zsteve: Fixed

examples/empirical_sinkhorn_div/script.jl
```diff
@@ -0,0 +1,94 @@
...
+# Define the loss function to minimise, where `x` specifies the locations of the Diracs in $\nu$. 
+function loss(x, ε)
+    C_μν = pairwise(SqEuclidean(), μ_spt', x')
+    C_ν = pairwise(SqEuclidean(), x')
+    return sinkhorn_divergence(μ, ν, C_μν, C_μ, C_ν, ε; maxiter=50)
+end
+# Set entropy regularisation parameter
+ε = 1.0;
+
+# Use ReverseDiff with a precompiled tape and Optim.jl to minimise $\nu \mapsto \operatorname{S}_{\varepsilon}(\mu, \nu)$. Note that this is problem is *not* convex, so we find a local minimium.
```
devmotion, line 57: I don't think this is valid - ReverseDiff with a compiled tape will only yield correct gradients if the function does not contain any branches or if statements. However, the number of iterations in the Sinkhorn algorithm is determined at run time and not fixed.

src/entropic/sinkhorn.jl
```diff
@@ -120,6 +120,19 @@ function check_convergence(solver::SinkhornSolver)
     )
 end
 
+function plan(u, v, K)
+    return K .* add_singleton(u, Val(2)) .* add_singleton(v, Val(1))
+end
+
+# dual objective 
+function obj(u, v, Kv, K, ε)
```
devmotion, line 128: Maybe use a more descriptive name?

devmotion:
```suggestion
function sinkhorn_dual_objective(u, v, Kv, K, ε)
```

src/entropic/sinkhorn.jl
```diff
@@ -120,6 +120,19 @@ function check_convergence(solver::SinkhornSolver)
     )
 end
 
+function plan(u, v, K)
```
devmotion, line 123:
```suggestion
function sinkhorn_plan(u, v, K)
```

src/entropic/sinkhorn.jl
```diff
@@ -209,3 +222,113 @@ function sinkhorn2(μ, ν, C, ε, alg::Sinkhorn; regularization=false, plan=noth
 
     return cost
 end
+
+function sinkhorn_loss(μ, ν, C, ε, alg::Sinkhorn; kwargs...)
+    return error(
+        "sinkhorn_loss is only implemented for alg::SinkhornGibbs. For other algorithms please use sinkhorn2",
+    )
+end
+
```
devmotion, line 231: Can we just integrate this in `sinkhorn2`? Or rename `sinkhorn2` to `sinkhorn_loss`?

src/entropic/sinkhorn.jl
```diff
@@ -209,3 +222,113 @@ function sinkhorn2(μ, ν, C, ε, alg::Sinkhorn; regularization=false, plan=noth
...
+"""
+    sinkhorn_divergence(μ::AbstractVecOrMat, ν::AbstractVecOrMat, C, ε, alg::Sinkhorn = SinkhornGibbs(), algμ = SymmetricSinkhornGibbs(), algν = SymmetricSinkhornGibbs(); regularization = nothing, plan = nothing, kwargs...)
```
devmotion, line 233: 92 characters per line :slightly_smiling_face:

src/entropic/sinkhorn.jl
```diff
@@ -209,3 +222,113 @@ function sinkhorn2(μ, ν, C, ε, alg::Sinkhorn; regularization=false, plan=noth
...
+Compute the Sinkhorn Divergence between finite discrete
```
devmotion, line 234:
```suggestion

Compute the Sinkhorn Divergence between finite discrete
```

src/entropic/sinkhorn.jl
```diff
@@ -209,3 +222,113 @@ function sinkhorn2(μ, ν, C, ε, alg::Sinkhorn; regularization=false, plan=noth
...
+\\operatorname{S}_{ε}(μ,ν) := \\operatorname{W}_{ε}(μ,ν)
+- \\frac{1}{2}(\\operatorname{W}_{ε}(μ,μ) + \\operatorname{W}_{ε}(ν,ν)),
+```
+where ``\\operatorname{W}_{ε}(μ,ν)``, ``\\operatorname{W}_{ε}(μ,μ)`` and
+``\\operatorname{W}_{ε}(ν,ν)`` are the entropically regularized optimal transport *cost*
+between `(μ,ν)`, `(μ,μ)` and `(ν,ν)`, respectively, defined as 
+```math
+\\operatorname{W}_{ε}(μ, ν) = \\langle C, γ^\\star \\rangle
+```
+where ``γ^\\star`` is the entropy-regularised transport plan between `μ` and `ν`.
```
devmotion, line 258: Can we shorten this explanation? Seems a bit too elaborate for my taste, IMO it is sufficient to use `OT_eps` (as we always do) and say that one case includes the regularization term as Feydy et al. and the other doesn't as Genevay et al.

src/entropic/sinkhorn.jl
```diff
@@ -209,3 +222,113 @@ function sinkhorn2(μ, ν, C, ε, alg::Sinkhorn; regularization=false, plan=noth
...
+    μ::AbstractVecOrMat,
+    ν::AbstractVecOrMat,
+    C,
+    ε,
+    alg::Sinkhorn=SinkhornGibbs(),
+    algμ::Sinkhorn=SymmetricSinkhornGibbs(),
+    algν::Sinkhorn=SymmetricSinkhornGibbs();
+    regularization=true,
+    plan=nothing,
+    kwargs...,
```
devmotion, line 277: Can we just pass along `kwargs...` and handle everything in the other functions? Possibly set `regularization = true` if we want to use it as default:
```suggestion
    regularization=true,
    kwargs...,
```

src/entropic/sinkhorn.jl
```diff
@@ -209,3 +222,113 @@ function sinkhorn2(μ, ν, C, ε, alg::Sinkhorn; regularization=false, plan=noth
...
+)
+    OTμν, OTμ, OTν = if (regularization == true) && (plan === nothing)
+        sinkhorn_loss(μ, ν, C, ε, alg; kwargs...),
+        sinkhorn_loss(μ, C, ε, algμ; kwargs...),
+        sinkhorn_loss(ν, C, ε, algν; kwargs...)
+    else
+        sinkhorn2(μ, ν, C, ε, alg; plan=plan, regularization=false, kwargs...),
+        sinkhorn2(μ, C, ε, algμ; plan=nothing, regularization=false, kwargs...),
+        sinkhorn2(ν, C, ε, algν; plan=nothing, regularization=false, kwargs...)
+    end
```
devmotion, line 287: Again, can we handle this in the other functions?
```suggestion
    OTμν = sinkhorn2(μ, ν, C, ε, alg; regularization=regularization, kwargs...)
    OTμ = sinkhorn2(μ, C, ε, algμ; regularization=regularization, kwargs...)
    OTν = sinkhorn2(ν, C, ε, algν; regularization=regularization, kwargs...)
```

src/entropic/sinkhorn.jl
```diff
@@ -209,3 +222,113 @@ function sinkhorn2(μ, ν, C, ε, alg::Sinkhorn; regularization=false, plan=noth
...
+    return max.(0, OTμν .- (OTμ .+ OTν) / 2)
```
devmotion, line 288: Is `OTμν .- (OTμ .+ OTν) / 2` guaranteed to be non-negative even if the regularization terms are not included (as in Genevay et al.)?

src/entropic/sinkhorn.jl
```diff
@@ -209,3 +222,113 @@ function sinkhorn2(μ, ν, C, ε, alg::Sinkhorn; regularization=false, plan=noth
...
+        kwargs...,
+    )
+Compute the Sinkhorn Divergence between finite discrete
+measures `μ` and `ν` with respect to the precomputed cost matrices `Cμν`,
+`Cμμ` and `Cνν`, entropic regularization parameter `ε` and algorithm `alg`.
+The default algorithm is the `SinkhornGibbs`.
+A pre-computed optimal transport `plan` between `μ` and `ν` may be provided.
+See also: [`sinkhorn2`](@ref) 
+"""
+function sinkhorn_divergence(
```
devmotion, line 310: The other `sinkhorn_divergence` function should just be forwarded to this function, and the comments from above should be applied here.

src/entropic/sinkhorn.jl
```diff
@@ -209,3 +222,113 @@ function sinkhorn2(μ, ν, C, ε, alg::Sinkhorn; regularization=false, plan=noth
...
+    μ,
+    ν,
+    Cμν,
+    Cμ,
+    Cν,
+    ε,
+    alg::Sinkhorn=SinkhornGibbs(),
+    algμ::Sinkhorn=SymmetricSinkhornGibbs(),
+    algν::Sinkhorn=SymmetricSinkhornGibbs();
```
devmotion, line 319: I think it would be better to have a `SinkhornDivergence` algorithm that would wrap the individual algorithms, so that still only one algorithm is passed to the function.

devmotion (changes requested):

src/entropic/sinkhorn.jl
davibarreira: I don't think it's guaranteed due to numerics. Otherwise, yes, it's theoretically always greater or equal than zero.

src/entropic/sinkhorn.jl
```diff
@@ -290,16 +302,18 @@ function sinkhorn_divergence(
     alg::Sinkhorn=SinkhornGibbs(),
     algμ::Sinkhorn=SymmetricSinkhornGibbs(),
     algν::Sinkhorn=SymmetricSinkhornGibbs();
-    regularization=nothing,
+    regularization=true,
```
davibarreira, line 320: The default should be `false`, no?

src/entropic/sinkhorn.jl
devmotion: Sure, Feydy et al. showed non-negativity for the divergence that includes the regularization terms. But I haven't seen a statement or proof for the non-negativity of the divergence that does not include the regularization terms in Genevay et al.'s paper (they mentioned "positivity is yet to be proved" in the main text). Is it proven nowadays?

src/entropic/sinkhorn.jl
devmotion: Depending on which approach one follows :shrug: Feydy et al. included the regularization terms, Genevay et al. not (at least that's the way it's described in their papers).

src/entropic/sinkhorn.jl
davibarreira: Yes, it's proven. Actually, it's proven that this is an actual metric.

davibarreira: > Yes, it's proven. Actually, it's proven that this is an actual metric.

Let me find the paper and I'll share here.

davibarreira: Actually, the proof is in the Feydy's paper. It's Theorem 1.

davibarreira: I just saw that it seems that Feydy is incorporating the regularization term. But, I was pretty sure he didn't. I'll double check. Just a min.

examples/empirical_sinkhorn_div/script.jl
zsteve: Yes you're absolutely right. Makes sense to me but probably by that time I was copypasting from their tutorial.
A workaround for now would be to set `atol=rtol=0` and fix `maxiter = 50` or something. 
In the long run it might be useful to have an option to set if we want to avoid convergence checks and just run for fixed nb of iterations.

examples/empirical_sinkhorn_div/script.jl
```diff
@@ -0,0 +1,94 @@
...
+const compiled_loss_tape = ReverseDiff.compile(loss_tape)
+opt = with_logger(SimpleLogger(stderr, Logging.Error)) do
+    optimize(
+        x -> loss(x, ε),
+        (∇, x) -> ReverseDiff.gradient!(∇, compiled_loss_tape, x),
+        ν_spt,
+        GradientDescent(),
+        Optim.Options(; iterations=10, g_tol=1e-6, show_trace=true),
+    )
+end
```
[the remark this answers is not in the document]
zsteve: Right, makes sense!

src/entropic/sinkhorn.jl
zsteve: This was what I was hoping to accomplish initially. However, the considerations I had in mind were:
- `sinkhorn2` currently computes the transport plan matrix, whereas `sinkhorn_loss` only uses the dual variables `(u, v)`. So the internal implementations would need to be diffetagrent.
- We can only use the `sinkhorn_dual_obj` method when `regularization = true`. It would be nice to pass tags such as `DualObjective(), Cost()` etc and dispatch to different methods. 

If we want to handle it in `sinkhorn2`, then I think the quick fix will be to handle the special case `regularization == true && plan === nothing` inside. I guess both solutions for now are quite ugly.

src/entropic/sinkhorn.jl
zsteve: I do wonder whether this could cause issues with differentiability... however I guess the inputs that would fall afoul of `max` would be vanishingly few.

src/entropic/sinkhorn.jl
zsteve: Thanks for pointing this out. Previously there was an implementation difference hence it's still hanging around.

src/entropic/sinkhorn.jl
zsteve: I'll implement something like `SinkhornDivergence(SinkhornGibbs(), SymmetricSinkhornGibbs(), SymmetricSinkhornGibbs())`

src/entropic/sinkhorn.jl
davibarreira: I do believe that it's always positive. But indeed it's not proven. I'll try to figure it out, it does not look very complicated.

examples/empirical_sinkhorn_div/Project.toml
```diff
@@ -0,0 +1,10 @@
+[deps]
+Distances = "b4f34e82-e78d-54a5-968a-f98e89d6e8f7"
```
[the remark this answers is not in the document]
zsteve: Done!

examples/empirical_sinkhorn_div/script.jl
```diff
@@ -0,0 +1,94 @@
+# # Sinkhorn divergences 
+# 
+# In this tutorial we provide a minimal example for using the Sinkhorn divergence as a loss function [FSV+19] on empirical distributions. 
+# [FSV+19]: Feydy, Jean, et al. "Interpolating between optimal transport and MMD using Sinkhorn divergences." The 22nd International Conference on Artificial Intelligence and Statistics. PMLR, 2019.
+# 
+# While entropy-regularised optimal transport $\operatorname{OT}_{\varepsilon}(\cdot, \cdot)$ is commonly used as a loss function, it suffers from a problem of *bias*: namely that $\nu \mapsto \operatorname{OT}_{\varepsilon}(\mu, \nu)$ is *not* minimised at $\nu = \mu$. 
+#
+# Feydy et al. fix this problem by introducing the *Sinkhorn divergence* between two measures $\mu$ and $\nu$, defined as
```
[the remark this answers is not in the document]
zsteve: Yes indeed. I've amended the sentence now to give fair credit to the Genevay paper.

examples/empirical_sinkhorn_div/script.jl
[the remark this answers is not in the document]
zsteve: With some more testing, compiling first the tape improves performance significantly. I'm leaving the compiled tape here but have set `rtol = atol = 0` and `maxiter = 50`, ensuring that exactly 50 iterations are done each time. Again, would be open to working out a modification to the API to allow for fixed # of iterations in a more clear way.

src/entropic/sinkhorn.jl
zsteve: Done

src/entropic/sinkhorn.jl
zsteve: Done!

src/entropic/sinkhorn.jl
zsteve: I don't really know/have a preference. Although with the current "OT" notation we've used I'd say `regularization = true` is most consistent.

davibarreira: I think we could go with true, but I think we should emphasize it, since the default for sinkhorn is false.

zsteve: Oops sorry my mistake, I forgot that sinkhorn uses false by default. Probably best to go with regularization=false. @devmotion ?

devmotion: Hmm I'm not sure. I agree that it would be more consistent to use `regularization=false` (i.e., to not specify it at all it seems). But then we would deviate from POT and would not implement the definition in Feydy et al. but Genevay et al.

examples/empirical_sinkhorn_div/script.jl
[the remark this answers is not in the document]
devmotion: A compiled tape usually improves performance significantly. In my opinion we should explain more clearly why one has to be careful and why it leads to correct values here.

src/entropic/sinkhorn_gibbs.jl
```diff
@@ -119,12 +119,21 @@ function sinkhorn2(
...
+    # return loss
+    cache = solver.cache
+    return sinkhorn_dual_objective(cache.u, cache.v, cache.Kv, cache.K, solver.eps)
+end
+
 # interface
 
 prestep!(::SinkhornSolver{SinkhornGibbs}, ::Int) = nothing
 
 function plan(solver::SinkhornSolver{SinkhornGibbs})
```
devmotion, line 136: Rename this function to `sinkhorn_plan` as well?

src/entropic/sinkhorn.jl
devmotion: > If we want to handle it in `sinkhorn2`, then I think the quick fix will be to handle the special case `regularization == true && plan === nothing` inside as an if-else block. I guess both solutions for now are quite ugly.

I think we should do this nevertheless. IMO it will be more readable and a cleaner API if there is only one function. I assume we could also use a similar approach for the case `regularization = false` and `plan = nothing` and avoid having to instantiate the plan?

examples/empirical_sinkhorn_div/script.jl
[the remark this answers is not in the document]
zsteve: Makes sense! I'll add a sentence about it.

src/entropic/sinkhorn.jl
zsteve: Yeah sounds good. One additional consideration is that it should only be used for `SinkhornGibbs` algorithm. I've implemented this now.

src/entropic/sinkhorn.jl
zsteve: Latest commit sets `regularization = false` now. Probably not a huge problem to deviate from POT.

src/entropic/sinkhorn_divergence.jl
```diff
@@ -0,0 +1,97 @@
...
+    ε,
+    alg::SinkhornDivergence=SinkhornDivergence(
+        SinkhornGibbs(), SymmetricSinkhornGibbs(), SymmetricSinkhornGibbs()
+    );
+    regularization=false,
+    kwargs...,
+)
+    OTμν = sinkhorn2(μ, ν, Cμν, ε, alg.algμν; regularization=regularization, kwargs...)
+    OTμμ = sinkhorn2(μ, Cμ, ε, alg.algμμ; regularization=regularization, kwargs...)
+    OTνν = sinkhorn2(ν, Cν, ε, alg.algνν; regularization=regularization, kwargs...)
```
devmotion, line 95: We don't have to specify `regularization=false` since it is the default in the downstream algorithms:
```suggestion
        kwargs...,
    )

Compute the Sinkhorn Divergence between finite discrete
measures `μ` and `ν` with respect to the precomputed cost matrices `Cμν`,
`Cμμ` and `Cνν`, entropic regularization parameter `ε` and algorithm `alg`.

A pre-computed optimal transport `plan` between `μ` and `ν` may be provided.

See also: [`sinkhorn2`](@ref), [`sinkhorn_divergence`](@ref)
"""
function sinkhorn_divergence(
    μ,
    ν,
    Cμν,
    Cμ,
    Cν,
    ε,
    alg::SinkhornDivergence=SinkhornDivergence(
        SinkhornGibbs(), SymmetricSinkhornGibbs(), SymmetricSinkhornGibbs()
    );
    kwargs...,
)
    OTμν = sinkhorn2(μ, ν, Cμν, ε, alg.algμν; kwargs...)
    OTμμ = sinkhorn2(μ, Cμ, ε, alg.algμμ; kwargs...)
    OTνν = sinkhorn2(ν, Cν, ε, alg.algνν; kwargs...)
```

src/entropic/symmetric.jl
```diff
@@ -0,0 +1,149 @@
...
+    # compute optimal transport plan
+    γ = sinkhorn_plan(solver)
+
+    return γ
+end
+
+function sinkhorn2(
+    μ, C, ε, alg::SymmetricSinkhornGibbs; regularization=false, plan=nothing, kwargs...
+)
+    cost = if (regularization == true) && (plan === nothing)
```
devmotion, line 127:
```suggestion
    cost = if regularization && plan === nothing
```

src/entropic/sinkhorn_gibbs.jl
```diff
@@ -119,12 +119,41 @@ function sinkhorn2(
...
+        else
+            # check dimensions
+            checksize(μ, ν, C)
+            size(plan) == size(C) || error(
+                "optimal transport plan `plan` and cost matrix `C` must be of the same size",
+            )
+            plan
+        end
+        cost = sinkhorn_cost_from_plan(γ, C, ε; regularization=regularization)
+        cost
```
devmotion, line 147:
```suggestion
        sinkhorn_cost_from_plan(γ, C, ε; regularization=regularization)
```

src/entropic/sinkhorn_gibbs.jl
```diff
@@ -119,12 +119,41 @@ function sinkhorn2(
     )
 end
 
+# spceialised sinkhorn2 for SinkhornGibbs
+function sinkhorn2(
+    μ, ν, C, ε, alg::SinkhornGibbs; regularization=false, plan=nothing, kwargs...
+)
+    cost = if (regularization == true) && (plan === nothing)
```
devmotion, line 126:
```suggestion
    cost = if regularization && plan === nothing
```

src/entropic/symmetric.jl
```diff
@@ -0,0 +1,149 @@
...
+        γ = if plan === nothing
+            sinkhorn(μ, C, ε, alg; kwargs...)
+        else
+            size(plan) == size(C) || error(
+                "optimal transport plan `plan` and cost matrix `C` must be of the same size",
+            )
+            plan
+        end
+        cost = sinkhorn_cost_from_plan(γ, C, ε; regularization=regularization)
+        cost
```
devmotion, line 146:
```suggestion
        sinkhorn_cost_from_plan(γ, C, ε; regularization=regularization)
```

examples/empirical_sinkhorn_div/script.jl
```diff
@@ -0,0 +1,103 @@
...
+    C_μν = pairwise(SqEuclidean(), μ_spt', x')
+    C_ν = pairwise(SqEuclidean(), x')
+    return sinkhorn_divergence(
+        μ, ν, C_μν, C_μ, C_ν, ε; maxiter=50, atol=rtol = 0, regularization=true
+    )
+end
+# Set entropy regularisation parameter
+ε = 1.0;
+
+# Use ReverseDiff with a precompiled tape and Optim.jl to minimise $\nu \mapsto \operatorname{S}_{\varepsilon}(\mu, \nu)$. Note that this is problem is *not* convex, so we find a local minimium.
```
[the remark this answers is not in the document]
zsteve: It's not convex because we are optimising over the positions of the points, not the weights. E.g. we can move from one local minima to another by exchanging points.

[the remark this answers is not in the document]
devmotion: True, I missed that these points are fixed in the notation of Theorem 1 in the paper.

zsteve:
If everything looks good so far, will bump the version and merge once I get a chance. 
@devmotion  any idea what is going on with some of the failing actions? e.g. CI Julia 1/Ubuntu x64 fails with the following:
```
Run coverallsapp/github-action@master
Using lcov file: lcov.info
Error: Bad response: 405 <html>
<head><title>405 Not Allowed</title></head>
```
I'll look into it in the next few days since I'm a little busy rn, but wondering if there's an easy fix.

devmotion:
I think coveralls was down for maintenance.

PR #146, opened 2021-09-22
add sinkhorn_divergence to runtests

zsteve:
I am clearly a donkey and didn't include sinkhorn_divergence in `runtests.jl` (the tests all pass, though).
will merge directly

### DIFF
--- a/Project.toml
+++ b/Project.toml
@@ -4,7 +4,6 @@ authors = ["zsteve <stephenz@student.unimelb.edu.au>"]
 version = "0.3.17"
 
 [deps]
-Compat = "34da2185-b29b-5c13-b0c7-acf172513d20"
 ExactOptimalTransport = "24df6009-d856-477c-ac5c-91f668376b31"
 IterativeSolvers = "42fd0dbc-a981-5370-80f2-aaf504508153"
 LinearAlgebra = "37e2e46d-f89d-539d-b4ee-838fcccc9c8e"
@@ -13,7 +12,6 @@ NNlib = "872c559c-99b0-510c-b3b7-b6c96a88d5cd"
 Reexport = "189a3867-3050-52da-a836-e630ba90ab69"
 
 [compat]
-Compat = "2.2, 3"
 ExactOptimalTransport = "0.1"
 IterativeSolvers = "0.8.4, 0.9"
 LogExpFunctions = "0.2, 0.3"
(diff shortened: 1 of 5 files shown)


Project.toml
```diff
@@ -4,6 +4,7 @@ authors = ["zsteve <stephenz@student.unimelb.edu.au>"]
 version = "0.3.17"
 
 [deps]
+Compat = "34da2185-b29b-5c13-b0c7-acf172513d20"
```
devmotion, line 7: Can we just replace `eachcol` with `(view(x, :, i) for i in axes(x, 2))` (that's its definition)? Compat is a quite large package and it seems a bit strange to depend on it just for one line.

src/utils.jl
```diff
@@ -25,7 +25,7 @@ end
 dot_matwise(x::AbstractMatrix, y::AbstractArray) = dot_matwise(y, x)
 
 function dot_vecwise(x::AbstractMatrix, y::AbstractMatrix)
-    return [dot(u, v) for (u, v) in zip(eachcol(x), eachcol(y))]
+    Compat.@compat return [dot(u, v) for (u, v) in zip(eachcol(x), eachcol(y))]
```
devmotion, line 28: As mentioned, I think it would be better to not use Compat at all but rewrite the code. In any case, you don't have to use `@compat` - the macro is only useful if some Julia syntax has changed and has to be rewritten before it is compiled. Here it is sufficient to load Compat.

test/entropic/sinkhorn_divergence.jl
```diff
@@ -53,17 +53,17 @@ Random.seed!(100)
             ν = hcat([normalize!(f.(x; μ=randn(), σ=0.5), 1) for _ in 1:M]...)
             for reg in (true, false)
                 loss_batch = sinkhorn_divergence(μ, ν, C, ε; regularization=reg)
-                @test loss_batch ≈ [
+                @compat @test loss_batch ≈ [
```
devmotion, line 56: Same as above, I think it would be preferable to rewrite the code. In any case `@compat` is not needed.

Project.toml
zsteve: Yep fair enough.

src/utils.jl
zsteve: Yep, rewriting this too. My initial thoughts were that the `eachcol` notation was quite compact, but now that you mention it I agree it's not worth an additional dependency

PR #147, opened 2021-10-20
Example for 1D case of optimal transport

davibarreira:
Recently in Julia's Slack, a person came asking about the 1D Wasserstein distance, and I realized that we don't have a clear example in the docs. So I wrote one. The example has both a continuous case and a discrete case.

### DIFF
--- a/.github/workflows/Documentation.yml
+++ b/.github/workflows/Documentation.yml
@@ -15,7 +15,7 @@ jobs:
       - name: Install python
         uses: actions/setup-python@v2
         with:
-          python-version: '3.x'
+          python-version: '3.9'
           architecture: ${{ matrix.arch }}
       # Limitation of pip: https://pythonot.github.io/index.html#pip-installation
       - run: python -m pip install cython numpy


devmotion:
> I haven't written examples with `Literate.jl` before. I just created a new folder and followed a similar style to the other examples. Do I need to run a script to build the docs or does GitHub Actions takes care of it?

Yes, this is correct, one just has to use the same structure as the existing examples. The script `docs/make.jl` (which is also executed by the Github action) runs all examples in these subfolders automatically.

Currently, the example fails since the Project.toml (and I assume Manifest.toml) do not include the Distributions dependency: https://github.com/JuliaOptimalTransport/OptimalTransport.jl/pull/147/checks?check_run_id=3957270384#step:8:22 The Project.toml should list all dependencies of the example (including Literate and OptimalTransport; the latter you should add with a relative path, e.g., with `] dev ../..`: https://github.com/JuliaOptimalTransport/OptimalTransport.jl/blob/59223df8db5a6e776d87244b7e50eb57d60e89ae/examples/empirical_sinkhorn_div/Manifest.toml#L580). It seems you can simplify it a bit as well, e.g. you can remove PythonOT from the Project.toml here.

You can find some instructions regarding the examples here: https://github.com/JuliaGaussianProcesses/KernelFunctions.jl/tree/master/examples#add-a-new-example (we use the same setup)

devmotion:
Yeah, the installation of POT (the Python library, not the Julia package itself) is really unstable. There are quite many issues about it in their repo and it seems they plan to improve it in an upcoming release. I'll have a look within the next days if we can work around it somehow. In the worst case we could disable tests with Windows until it is fixed upstream.

Could you also build the docs locally and check it looks correct (maybe also post a screenshot)? The online preview does not work for PRs from forks (not our fault, it's a security feature from Github and checked in Documenter).

davibarreira:
image
image
image
image

zsteve:
Yes, the example looks fine. Thanks for adding this!

PR #158, opened 2021-12-22
Update dependencies and CI

devmotion:
This PR updates ExactOptimalTransport and Tulip dependencies and CI.

Fixes https://github.com/JuliaOptimalTransport/OptimalTransport.jl/pull/156, https://github.com/JuliaOptimalTransport/OptimalTransport.jl/pull/155, https://github.com/JuliaOptimalTransport/OptimalTransport.jl/pull/154.


devmotion:
Test failures are probably just random failures caused by the RNG changes in 1.7. I'll check it.

devmotion:
Hmm I can't reproduce the error locally, possibly only an architecture dependent floating point issue.

devmotion:
I have no idea what causes the OpenBLAS error with Julia 1.7 on Windows: https://github.com/JuliaOptimalTransport/OptimalTransport.jl/runs/4612478949?check_suite_focus=true#step:8:190 Interestingly, Julia nightly seems to work on Windows.

devmotion:
It seems the problem was caused by the OpenBLAS issues in base that were fixed in Julia 1.6.5 and 1.7.1.

Issue #151, opened 2021-10-29
Greenkhorn and screenkhorn - near-linear time sinkhorn


flyaflya:
Thx for the great package,  I have just been toying with it.  I noticed there are some newer near-linear time sinkhorn-based algorithms (Greenkhorn and screenkhorn) that are available in POT.  The link to the Greenkhorn paper is below.

Just curious if these are on the roadmap for OptimalTransport.jl?

https://arxiv.org/pdf/1705.09634.pdf

zsteve:
> Thx for the great package, I have just been toying with it. I noticed there are some newer near-linear time sinkhorn-based algorithms (Greenkhorn and screenkhorn) that are available in POT. The link to the Greenkhorn paper is below.
> 
> Just curious if these are on the roadmap for OptimalTransport.jl?
> 
> https://arxiv.org/pdf/1705.09634.pdf

@flyaflya glad to hear this package has been helpful! I don't think there's any reason why this should not be a candidate algorithm for inclusion. (Basically it will depend on the amount of free time people have xD). 
However, if you implement these algorithms as part of your own research, please feel more than welcome to update this issue or make a pull request! If you have existing code, am more than happy to help adapt it to fit into the package.

davibarreira:
@flyaflya #159

PR #159, opened 2022-01-16
Implementing the Greenkhorn

davibarreira:
This PR is related to #151 .
I've implemented the Greenkhorn algorithm which is a greedy version of the Sinkhorn algorithm. The method I've implemented is actually the one in POT, which is a bit different from the one in the original paper (https://arxiv.org/pdf/1705.09634.pdf).

The implementation needs improvement. I was not able to get it to work with AD and with the batch tests. I was not very involved in the coding of the original Sinkhorn algorithm, and had some difficulty getting around the whole `step!`, `solve!` and `cache` structure.

Another point. The iteration in the Greenkhorn algorithm only updates one row of `u` and `v` each time, thus, it needs many more iterations in order to converge compared to the original Sinkhorn. Some preliminar benchmarks showed that this implementation of the Greenkhorn is slower then the original Sinkhorn_Gibbs, which seems to contradict the claims in the paper. I believe the reason for this might be that the Sinkhorn implementation is very optimized  in the package compared to my version of Greenkhorn. Another possibility is that the Sinkhorn version of the paper was not very efficient ( if you read it [here](https://arxiv.org/pdf/1705.09634.pdf), they present the Sinkhron algorithm which computer `diagm(u) K diagm(v)` in each iteration).

I've compared the results from my algorithm against POT, and indeed it seems to be returning the exact same result each iteration, i.e. it seems that the Greenkhorn implementation is correct but not optimal.

### DIFF
--- a/src/entropic/greenkhorn.jl
+++ b/src/entropic/greenkhorn.jl
@@ -59,7 +59,6 @@ function build_cache(
         dv = reshape(sum(G, dims=1), size(ν)) - ν
     # end
 
-    println(size(G), size(du), size(dv))
 
     return GreenkhornCache(u, v, K, Kv, G, du, dv)
 end
--- a/src/entropic/greenkhorn.jl
+++ b/src/entropic/greenkhorn.jl
@@ -1,0 +1,114 @@
+# Greenkhorn is a greedy version of the Sinkhorn algorithm
+# This method is from https://arxiv.org/pdf/1705.09634.pdf
+# Code is based on implementation from package POT
+
+
+"""
+    Greenkhorn()
+
+Greenkhorn is a greedy version of the Sinkhorn algorithm.
+"""
+struct Greenkhorn <: Sinkhorn end
+
+struct GreenkhornCache{U,V,KT}
+    u::U
+    v::V
+    K::KT
+    Kv::U
+    G::KT
+    du::U
+    dv::V
+end
+
+Base.show(io::IO, ::Greenkhorn) = print(io, "Greenkhorn algorithm")
+
+function build_cache(
+    ::Type{T},
+    ::Greenkhorn,
+    size2::Tuple,
+    μ::AbstractVecOrMat,
+    ν::AbstractVecOrMat,
+    C::AbstractMatrix,
+    ε::Real,
+) where {T}
+    # compute Gibbs kernel (has to be mutable for ε-scaling algorithm)
+    K = similar(C, T)
+    @. K = exp(-C / ε)
+
+    # create and initialize dual potentials
+    u = similar(μ, T, size(μ, 1), size2...)
+    v = similar(ν, T, size(ν, 1), size2...)
+    fill!(u, one(T)/size(μ, 1))
+    fill!(v, one(T)/size(ν, 1))
+
+    # G = sinkhorn_plan(u, v, K)
+    G = diagm(u) * K * diagm(v) 
+
+    Kv = similar(u)
+
+    # improve this!
+    # du = similar(u)
+    # fill!(du, 1.0)
+    # dv = similar(v)
+    du = sum(G', dims=1)[:] - μ
+    dv = sum(G', dims=2)[:] - ν
+
+    return GreenkhornCache(u, v, K, Kv, G, du, dv)
+end
+
+prestep!(::SinkhornSolver{Greenkhorn}, ::Int) = nothing
+
+function init_step!(solver::SinkhornSolver{<:Greenkhorn})
+    return A_batched_mul_B!(solver.cache.Kv, solver.cache.K, solver.cache.v)
+end
+
+function step!(solver::SinkhornSolver{<:Greenkhorn}, iter::Int)
+    μ = solver.source
+    ν = solver.target
+    cache = solver.cache
+    u = cache.u
+    v = cache.v
+    K = cache.K
+    G = cache.G
+    Δμ= cache.du
+    Δν= cache.dv
+
+    ρμ = abs.(Δμ + μ .* log.(μ ./ sum(G', dims=1)[:]))
+    ρν = abs.(Δν + ν .* log.(ν ./ sum(G', dims=2)[:]))
+    
+    i₁ = argmax(ρμ)
+    i₂ = argmax(ρν)
+
+    if ρμ[i₁]> ρν[i₂]
+        old_u = u[i₁]
+        u[i₁] = μ[i₁]/ (K[i₁,:] ⋅ v)
+        Δ = u[i₁] - old_u
+        G[i₁, :] = u[i₁] * K[i₁,:] .* v
+        Δμ[i₁] = u[i₁] * (K[i₁,:] ⋅ v) - μ[i₁]
+        Δν = Δν .+  Δ .* K[i₁,:] .* v
+    else
+        old_v = v[i₂]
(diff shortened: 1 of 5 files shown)


src/entropic/greenkhorn.jl
```diff
@@ -0,0 +1,114 @@
+# Greenkhorn is a greedy version of the Sinkhorn algorithm
+# This method is from https://arxiv.org/pdf/1705.09634.pdf
+# Code is based on implementation from package POT
```
devmotion, line 3: It would be helpful to describe what the differences are (if there are any apart from implementation details).

davibarreira: Yeah, there are. The paper implementation is actually just a couple of lines commented out. I'll point out in the code.

davibarreira: Just realized that it's already in the code. Inside the `step!` function.

src/entropic/greenkhorn.jl
```diff
@@ -0,0 +1,114 @@
...
+
+Greenkhorn is a greedy version of the Sinkhorn algorithm.
+"""
+struct Greenkhorn <: Sinkhorn end
+
+struct GreenkhornCache{U,V,KT}
+    u::U
+    v::V
+    K::KT
+    Kv::U #placeholder
```
devmotion, line 17: Why "placeholder"?

src/entropic/greenkhorn.jl
```diff
@@ -0,0 +1,114 @@
...
+    Δν= cache.dv
+
+    # The implementation in POT does not compute `μ .* log.(μ ./ sum(G', dims=1)[:])`
+    # or `ν .* log.(ν ./ sum(G', dims=2)[:])`. Yet, this term is present in the original
+    # paper, where it uses ρ(a,b) = b - a + a log(a/b).
+    # ρμ = abs.(Δμ + μ .* log.(μ ./ sum(G', dims=1)[:]))
+    # ρν = abs.(Δν + ν .* log.(ν ./ sum(G', dims=2)[:]))
+    
+    i₁ = argmax(abs.(Δμ))
+    i₂ = argmax(abs.(Δν))
```
devmotion, line 88:
```suggestion
    Δμ_max, Δμ_max_idx = findmax(abs, Δμ)
    Δν_max, Δν_max_idx = findmax(abs, Δν)
```

src/entropic/greenkhorn.jl
```diff
@@ -0,0 +1,114 @@
...
+
+    # if ρμ[i₁]> ρν[i₂]
+    if abs(Δμ[i₁]) > abs(Δν[i₂])
```
devmotion, line 91:
```suggestion
    if Δμ_max > Δν_max
```

src/entropic/greenkhorn.jl
```diff
@@ -0,0 +1,114 @@
...
+        old_u = u[i₁]
```
devmotion, line 92: Also has to be changed for batch support it seems.
```suggestion
        old_u = u[Δμ_max_idx]
```

src/entropic/greenkhorn.jl
```diff
@@ -0,0 +1,114 @@
...
+        u[i₁] = μ[i₁]/ (K[i₁,:] ⋅ v)
```
devmotion, line 93:
```suggestion
        u[Δμ_max_idx] = μ[Δμ_max_idx] / dot(K[Δμ_max_idx, :], v)
```
It would be better to select columns instead of rows and to use views. Julia uses column major order, so a column is close in memory and hence accessing columns is faster.

src/entropic/greenkhorn.jl
```diff
@@ -0,0 +1,114 @@
...
+        Δ = u[i₁] - old_u
```
devmotion, line 94:
```suggestion
        Δ = u[Δμ_max_idx] - old_u
```

src/entropic/greenkhorn.jl
```diff
@@ -0,0 +1,114 @@
...
+        G[i₁, :] = u[i₁] * K[i₁,:] .* v
```
devmotion, line 95: Again, better to work with columns than with rows. Also some unnecessary allocations here:
```suggestion
        G[Δμ_max_idx, :] .= u[Δμ_max_idx] .* K[Δμ_max_idx, :] .* v
```

src/entropic/greenkhorn.jl
```diff
@@ -0,0 +1,114 @@
...
+        Δμ[i₁] = u[i₁] * (K[i₁,:] ⋅ v) - μ[i₁]
```
devmotion, line 96: Same here.

src/entropic/greenkhorn.jl
```diff
@@ -0,0 +1,114 @@
...
+        @. Δν = Δν +  Δ * K[i₁,:] * v
```
devmotion, line 97: And here.

src/entropic/greenkhorn.jl
```diff
@@ -0,0 +1,114 @@
...
+    else
```
devmotion, line 98: Same comments in the second branch.

src/entropic/greenkhorn.jl
davibarreira: This is a partial solution. Without this `Kv`, I got an error. It seems that  the `Sinkhorn` structs further on require it. I could not find out how to get rid of it without changing the code for Sinkhorn.

src/entropic/greenkhorn.jl
davibarreira: True! I'll try to come up with something.

src/entropic/greenkhorn.jl
davibarreira: What is the unnecessary allocation?

src/entropic/greenkhorn.jl
devmotion: I mean you use it explicitly below for checking convergence?

src/entropic/greenkhorn.jl
devmotion: Oh there are multiple unnecessary allocations.

First of all, `K[i₁, :]` creates, i.e., allocates, a row vector. Then `u[i₁] * K[i₁, :]` scales it and allocates a new row vector. And finally `u[i₁] * K[i₁,:] .* v` multiplies the entries of `u[i₁] * K[i₁,:]` elementwise with `v` and allocates yet another row vector.

Whereas the alternative suggestion allocates only `K[i₁, :]` (could be avoided by using a `view`) and then fuses all multiplications and writes the result directly to `G` without allocating any other row vector.

src/entropic/greenkhorn.jl
davibarreira: Yeah, it's been used to check the convergence. But I think the convergence check might be done another way more efficiently. But I was having trouble getting it to work with the existing "api" for Sinkhorn, so I updated `Kv` and used the convergence verification already in place.

You are right that, as is, `Kv` is not actually just a placeholder, since it's been used in the convergence.

src/entropic/greenkhorn.jl
davibarreira: Thanks for the answer, and sorry for the bad code. I'm still very crude with code optimization.

PR #160, opened 2022-01-17
Symmetric quad [DRAFT]

zsteve:
The semi-smooth Newton algorithm of [Lorenz et al. 2019] can be specialised to the case of symmetric inputs, much like the approach already used for symmetric Sinkhorn algorithm in the already implemented Sinkhorn divergence. This problem can be solved faster than the non-symmetric case.

Reference:
Quadratically regularized optimal transport. DA Lorenz, P Manns, C Meyer - Applied Mathematics & Optimization, 2019


davibarreira:
Are we supposed to review this?  The "[Draft]" is throwing me off.

PR #165, opened 2022-03-08
Entropy-regularised Gromov-Wasserstein

zsteve:


.github/workflows/CI.yml
```diff
@@ -20,7 +20,7 @@ jobs:
     strategy:
       matrix:
         version:
-          - '1.6'
+          - '1.8'
```
devmotion, line 23: This is the same as '1' (see https://github.com/JuliaOptimalTransport/PythonOT.jl/pull/21). IMO it's important to test the oldest Julia version that we want to support, the stable version, and possibly if not included already Julia LTS. So I suggest reverting this change:
```suggestion
          - '1.6'
```

gpu/Manifest.toml
```diff
@@ -0,0 +1,129 @@
+# This file is machine-generated - editing it directly is not advised
```
devmotion, line 1: Did you commit this file accidentally?

gpu/Project.toml
```diff
@@ -0,0 +1,2 @@
+[deps]
```
devmotion, line 1: Same, should this file be here?

src/gromov.jl
```diff
@@ -0,0 +1,91 @@
...
+    plan_prev = similar(C)
+    plan_prev .= plan
+    norm_plan = sum(plan)
+
+    _atol = atol === nothing ? 0 : atol
+    _rtol = rtol === nothing ? (_atol > zero(_atol) ? zero(T) : sqrt(eps(T))) : rtol
+
+    function get_new_cost!(C, plan, tmp, Cμ, Cν)
+        A_batched_mul_B!(tmp, Cμ, plan)
+        return A_batched_mul_B!(C, tmp, -4Cν)
```
devmotion, line 59: The multiplication with `-4` introduces additional allocations. I wonder if this could be avoided, e.g., by updating in-place or some additional cache.

src/gromov.jl
```diff
@@ -0,0 +1,91 @@
...
+        solver = build_solver(μ, ν, C, ε, alg.alg_step; kwargs...)
+        solve!(solver)
+        # compute optimal transport plan
+        plan = sinkhorn_plan(solver)
+
+        to_check_step -= 1
+        if to_check_step == 0 || iter == maxiter
+            # reset counter
+            to_check_step = check_convergence
+            isconverged = sum(abs, plan - plan_prev) < max(_atol, _rtol * norm_plan)
```
devmotion, line 80: `norm_plan` is never updated it seems but always set to `sum(plan)` of the initial randomly initialized `plan`?

devmotion: Maybe also avoid allocations here by writing:
```suggestion
            plan_prev .-= plan # used as a temporary array here to reduce allocations
            isconverged = sum(abs, plan_prev) < max(_atol, _rtol * norm_plan)
```

test/Project.toml
```diff
@@ -0,0 +1,2 @@
+[deps]
```
devmotion, line 1: This file should not be needed, remove?

test/runtests.jl
```diff
@@ -36,10 +36,14 @@ const GROUP = get(ENV, "GROUP", "All")
...
         end
+
+        @safetestset "Gromov-Wasserstein OT" begin
+            include("gromov.jl")
+        end
     end
 
     # CUDA requires Julia >= 1.6
-    if (GROUP == "All" || GROUP == "GPU") && VERSION >= v"1.6"
+    if (GROUP == "All" || GROUP == "GPU") && VERSION >= v"1.8"
```
devmotion, line 46: Why is 1.8 needed here?

test/runtests.jl
```diff
@@ -1,5 +1,5 @@
 using OptimalTransport
-using Pkg: Pkg
+using Pkg
```
devmotion, line 2: I think this change is not needed?
```suggestion
using Pkg: Pkg
```

gpu/Manifest.toml
zsteve: Looks like it woops

gpu/Project.toml
zsteve: Not sure how the gpu/ directory got there (and I didn't notice it somehow). Thanks, removed.

test/runtests.jl
zsteve: Sorry, still in the middle of playing around to see why the buildkite CI fails.

test/runtests.jl
zsteve: as above. something odd happening with the CI

src/gromov.jl
zsteve: > `norm_plan` is never updated it seems but always set to `sum(plan)` of the initial randomly initialized `plan`?

The initial plan is taken to be the independent coupling and here we only consider the balanced problem, so `norm_plan` should not change. I agree however this is a special case of the unbalanced problem where it would not be constant.

zsteve: > Maybe also avoid allocations here by writing:

Good catch, done

src/gromov.jl
zsteve: Dealt with now (in-place scaling of tmp)

PR #170, opened 2022-08-15
Unbalanced divergence

zsteve:
Implementation of unbalanced Sinkhorn divergence (see https://arxiv.org/pdf/1910.12958.pdf, and a short summary here https://thibsej.github.io/files/poster_neurips_2019.pdf), when the marginal constraints are KL divergence.
In particular uses a fast fixed point iteration for the symmetric terms.


zsteve:
Oops just realized there are some emacs cache files. Will remove in next commit

PR #171, opened 2022-08-21
remove deprecation tests in attempt to fix CI

zsteve:


zsteve:
It looks like for certain versions of Julia on Windows, POT/PythonOT failing to converge triggers some kind of memory error. I've not been able to reproduce this on any of my Linux machines. This is likely something upstream to do with PyCall.

zsteve:
looks like the ReadOnlyMemoryError happens regardless of the convergence warnings. Will need to investigate further.

PR #185, opened 2024-05-07
Update Compat Bounds for NNlib

jacobadenbaum:
I've updated the compat bounds for NNlib to allow v0.9

I've also fixed an issue with the GPU tests where the test environment was erroring because it required an old version of CUDA and NNlibCUDA (which has been deprecated in favor of a CUDA extension in NNlib).  

Please note: I have not been able to verify that the GPU tests run, since I do not have access to CUDA on my development machine.

### DIFF
--- a/Project.toml
+++ b/Project.toml
@@ -15,7 +15,7 @@ Reexport = "189a3867-3050-52da-a836-e630ba90ab69"
 ExactOptimalTransport = "0.1, 0.2"
 IterativeSolvers = "0.8.4, 0.9"
 LogExpFunctions = "0.2, 0.3"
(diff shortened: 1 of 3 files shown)


jacobadenbaum:
I'll note: although it looks like the CI tests for this are failing, I don't think that has anything to do with this PR.  It would be great if a maintainer could look at the CI setup for this to figure out what the problem is, since its not something that I think I can resolve here

davibarreira (commented):
@devmotion and @zsteve , this PR is just updating the version. I haven't worked on GPU stuff, so I haven't figured out why the error in the testing.

PR #186, opened 2024-07-06
Set PYTHON env variable in buildkite runs

devmotion:


matbesancon:
is this one still necessary now that buildkite seems to work?

PR #187, opened 2024-11-15
Adding Functionality for Linearized Optimal Transport Computations

dcgentile:
Hi! Thanks for writing a great package. I'm currently utilizing it for a research project where I need to compute O(n^2) transport costs, and to alleviate some of the computational cost, I am approximating the distance via a statistical estimator for the linearization of the Wasserstein distance. To get that estimator, one only needs access to the Sinkhorn potentials, but as far as I could tell, there wasn't an straightforward way to access those as the package currently exists. Therefore, I added one function that returns the Sinkhorn potentials, rather than the plan or the cost, and another function which returns a function that evaluates the estimator of transport map.


matbesancon:
@dcgentile are you still interested in finishing this PR?

PR #193, opened 2025-08-02
Fix POT link in README.md

matbesancon:
The link was pointing to the book, not to the POT package


devmotion (approved):
A good catch! A bit surprising given that the correct link is used further down in the README 😄

